Issue #2, opened 2021-12-03
Migration of File System Access issues applicable to File System


annevk:
Any issues that apply to the text in this standard filed at https://github.com/WICG/file-system-access/issues should be moved over.

janvarga:
This is already being discussed here: https://github.com/WICG/file-system-access/issues/323

People are actually requesting fully sync OPFS surface.

a-sully:
I believe that's the last of the issues relevant to this standard! All migrated issues have been closed on the WICG side with a comment redirecting any further conversation here.

annevk:
Thanks @a-sully! Much appreciated.

Issue #3, opened 2021-12-03
Queue tasks


annevk:
While preparing #1 I noticed that many promises are resolved and rejected from in parallel steps. Instead this has to be done from tasks. https://html.spec.whatwg.org/#queue-a-global-task seems most appropriate here, not entirely sure if this needs a new task source. Perhaps we should have one for storage in general?

a-sully:
I'm supportive of this!

Seeing as AccessHandles is supposed to be the fastest storage primitive ever exposed to the web, does it makes sense to provide the Access Handles API its own task source? (in addition to one for storage more broadly)

annevk:
Yeah, I think that could be reasonable.

cc @smaug---- @domenic

a-sully:
Revisiting this... All methods on the SyncAccessHandle are now sync (as per #7), so the argument made in my above comment about wanting a separate task source for this API is somewhat less relevant (although we'd like to eventually support an async version of SyncAccessHandles which is similarly performant - see #41).

It seems like we should have a new task source. @annevk do you have a preference for whether it's for this API or for storage in general?

annevk:
I think ideally we consider which tasks need to happen in order across storage specifications and which can go out-of-order. And then put those in separate sources?

a-sully:
Hmm good point. Thinking about this a bit more.... Writes using any storage API (including this one) can be rejected if the site is out of quota. So do all storage APIs need to use same task source for writes, since [they all interact with the Quota Management API](https://www.w3.org/TR/quota-api/#accessing-storagequota-interface)?

There are a couple of other [[1](https://www.w3.org/TR/IndexedDB/#request-source), [2](https://www.w3.org/TR/FileAPI/#blobreader-task-source)] storage-related task sources that I'm aware of, but they both queue operations that are read-only (i.e. don't affect quota)

annevk:
That's a good point too.

So, the tasks are for informing the event loop about the result of a storage operation. So yes, write storage operations need to be in order (at least per site), but reporting whether that succeeded or not back to JavaScript (this involves tasks) doesn't necessarily have to be in the same order, but we might as well stick to a single task source for that I think.

a-sully:
I was about to file an issue in whatwg/storage but stumbled upon https://github.com/whatwg/storage/issues/89. I don't personally understand much of the discussion happening on that issue - is there more to it than defining a new task source for write operations in the storage spec and then pointing all the storage API specs, including this one, to it?

annevk:
[...]

PR #9, opened 2022-03-07
Add FileSystemHandle.remove method

a-sully:
_Migrated from https://github.com/WICG/file-system-access/pull/283_ (see #2)

Addresses https://github.com/WICG/file-system-access/issues/214

Adds the ability for a file or directory to delete itself. See the proposal: https://github.com/whatwg/fs/blob/main/proposals/Remove.md

Currently, it is not possible to remove a file or directory given its handle. You must obtain the handle of the parent directory and call `FileSystemDirectoryHandle.removeEntry()`.

This would enable the common use case where you obtain a file handle from `showSaveFilePicker()`, but then decide you don't want to save after all, and delete the file, or clearing an entire bucket file system.

   * Chrome
   * …
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://github.com/web-platform-tests/wpt/blob/master/file-system-access/script-tests/FileSystemBaseHandle-remove.js
   * Chrome: https://crbug.com/1114923
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/fs/9.html" title="Last updated on Jul 17, 2023, 1:18 AM UTC (a3797f6)">Preview</a> | <a href="https://whatpr.org/fs/9/bee4f50...a3797f6.html" title="Last updated on Jul 17, 2023, 1:18 AM UTC (a3797f6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -496,13 +496,16 @@ The <dfn method for=FileSystemHandle>remove(|options|)</dfn> method steps are:
        If that throws an exception, [=/reject=] |result| with that exception
        and abort these steps.
 
-         Note: If |options|["{{FileSystemRemoveOptions/recursive}}"] is true,
-         the removal can fail non-atomically. Some files or directories might
-         have been removed while other files or directories still exist.
+       Note: If |options|["{{FileSystemRemoveOptions/recursive}}"] is true,
+       the removal can fail non-atomically. Some files or directories might
+       have been removed while other files or directories still exist.
 
-         Issue(11): Better specify what possible exceptions this could throw.
+       Issue(11): Better specify what possible exceptions this could throw.
 
-    1. [=/Resolve=] |result| with `undefined`.
+    1. [=Enqueue the following steps=] to the [=file system queue=]:
+      1. [=file entry/lock/release|Release the lock=] on |entry|.
+      1. [=Queue a storage task=] with |global| to
+         [=/resolve=] |result| with `undefined`.
 
 1. Return |result|.
 
--- a/index.bs
+++ b/index.bs
@@ -332,6 +332,7 @@ interface FileSystemHandle {
   readonly attribute USVString name;
 
   Promise<boolean> isSameEntry(FileSystemHandle other);
+  Promise<undefined> remove(optional FileSystemRemoveOptions options = {});
 };
 </xmp>
 
@@ -413,6 +414,97 @@ The <dfn method for=FileSystemHandle>isSameEntry(|other|)</dfn> method steps are
      [=/resolve=] |p| with true.
   1. Otherwise [=/resolve=] |p| with false.
 1. Return |p|.
+
+</div>
+
+### The {{FileSystemHandle/remove()}} method ### {#api-filesystemhandle-remove}
+
+<div class="note domintro">
+  : await |handle| . {{FileSystemHandle/remove()|remove}}()
+  : await |handle| . {{FileSystemHandle/remove()|remove}}({ {{FileSystemRemoveOptions/recursive}}: false })
+  :: Deletes the [=file system entry=] [=locate an entry|locatable=] by
+     |handle|'s [=FileSystemHandle/locator=] from the underlying file system.
+
+     Attempting to delete a file or directory that does not exist is considered
+     success, while attempting to delete a non-empty directory will
+     result in a promise rejection.
+
+  : await |handle| . {{FileSystemHandle/remove()|remove}}({ {{FileSystemRemoveOptions/recursive}}: true })
+  :: Deletes the [=file system entry=] [=locate an entry|locatable=] by
+     |handle|'s [=FileSystemHandle/locator=] from the underlying file system.
+     If that [=file system entry=] is a [=directory entry=],
+     its contents will also be deleted recursively.
+
+     Attempting to delete a file or directory that does not exist is considered
+     success.
+</div>
+
+<div algorithm>
+The <dfn method for=FileSystemHandle>remove(|options|)</dfn> method steps are:
+
+1. Let |result| be [=a new promise=].
+1. Let |locator| be [=this=]'s [=FileSystemHandle/locator=].
+1. Let |global| be [=this=]'s [=relevant global object=].
+1. Let |isInABucketFileSystem| be true if
+   [=this=] [=FileSystemHandle/is in a bucket file system=];
+   otherwise false.
+1. [=Enqueue the following steps=] to the [=file system queue=]:
+  1. Let |entry| be the result of [=locating an entry=] given |locator|.
+  1. Let |accessResult| be the result of running |entry|'s
+     [=file system entry/request access=] given "`readwrite`".
+  1. If |accessResult|'s [=file system access result/permission state=]
+     is not "{{PermissionState/granted}}", [=queue a storage task=] with
+     |global| to [=/reject=] |result| with a {{DOMException}} of
+     |accessResult|'s [=file system access result/error name=] and
+     abort these steps.
+
+  1. If |entry| is `null`, [=queue a storage task=] with |global| to [=/reject=]
+     |result| with a "{{NotFoundError}}" {{DOMException}} and abort these steps.
+
+  1. Let |lockResult| be the result of [=file entry/lock/take|taking a lock=]
+     with "`exclusive`" on |entry|.
+
+     Issue(137): Support locking directory entries.
+
+  1. [=Queue a storage task=] with |global| to run these steps:
+    1. If |lockResult| is "`failure`", [=/reject=] |result| with a
+       "{{NoModificationAllowedError}}" {{DOMException}} and abort these steps.
+
+    1. If |entry| is a [=directory entry=]:
+      1. If |entry|'s [=directory entry/children=] [=set/is empty=] and
+         |options|["{{FileSystemRemoveOptions/recursive}}"] is false,
+         [=/reject=] |result| with an
+         "{{InvalidModificationError}}" {{DOMException}} and abort these steps.
+
+      1. Let |isTheRootOfABucketFileSystem| be true if
+         |isInABucketFileSystem| is true and
+         |entry|'s [=file system entry/parent=] is `null`;
+         otherwise false.
+      1. If |isTheRootOfABucketFileSystem| is true:
+        1. [=set/For each=] |child| of |entry|'s [=directory entry/children=]:
+          1. Attempt to remove |child| from the underlying file system.
+             If that throws an exception, [=/reject=] |result| with
+             that exception and abort these steps.
+        1. Set |entry|'s [=directory entry/children=] to an empty [=/set=].
+        1. [=/Resolve=] |result| with `undefined` and abort these steps.
+
+    1. If |entry|'s [=file system entry/parent=] is not `null`,
+       [=set/remove=] |entry| from |entry|'s [=file system entry/parent=]'s
+       [=directory entry/children=].
+
+    1. Attempt to remove |entry| from the underlying file system.
+       If that throws an exception, [=/reject=] |result| with that exception
+       and abort these steps.
+
+         Note: If |options|["{{FileSystemRemoveOptions/recursive}}"] is true,
+         the removal can fail non-atomically. Some files or directories might
+         have been removed while other files or directories still exist.
+
+         Issue(11): Better specify what possible exceptions this could throw.
+
+    1. [=/Resolve=] |result| with `undefined`.
+
+1. Return |result|.
 
 </div>
 


index.bs
```diff
@@ -148,6 +148,8 @@ interface FileSystemHandle {
   readonly attribute FileSystemHandleKind kind;
   readonly attribute USVString name;
 
+  Promise<void> remove(optional FileSystemRemoveOptions options = {});
```
annevk, line 151:
```suggestion
  Promise<undefined> remove(optional FileSystemRemoveOptions options = {});
```

annevk (commented):
You'll need to make your membership of the googlers GitHub organization public to appease the IPR bot.

index.bs
```diff
@@ -196,6 +198,67 @@ and return {{FileSystemHandleKind/"directory"}} otherwise.
...
+
+1. Let |result| be [=a new promise=].
+1. Run the following steps [=in parallel=]:
+  1. Let |entry| be <b>[=this=]</b>'s [=FileSystemHandle/entry=].
+  1. Let |permissionStatus| be the result of [=requesting access=]
+     given <b>[=this=]</b> and {{"readwrite"}}.
+     If that throws an exception, [=reject=] |result| with that exception and abort.
+  1. If |permissionStatus| is not {{PermissionState/"granted"}},
+     [=/reject=] |result| with a {{NotAllowedError}} and abort.
+
```
a-sully, line 234: We need to acquire an exclusive lock here (and also in `dir.removeEntry(file)`. I'm waiting for the language in https://github.com/whatwg/fs/pull/21 to mature so we can copy/paste the "grab an exclusive lock or fail" algorithm steps here.

Just a heads up that currently, we have [a WPT](https://github.com/web-platform-tests/wpt/blob/335c36d606c02daa4a35bd073d8c0bbd96a2fad3/fs/script-tests/FileSystemDirectoryHandle-removeEntry.js#L90-L103) (which was written before file locking was a thing) asserting that removal (via `removeEntry()`) succeeds when the file has an open writable. Since creating a writable grabs a shared lock, we'll have to update this WPT to make removal fail while any locks on the handle are held

index.bs
```diff
@@ -241,6 +243,66 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+<div algorithm>
+The <dfn method for=FileSystemHandle>remove(|options|)</dfn> method steps are:
+
+1. Let |result| be [=a new promise=].
+1. Run these steps [=in parallel=]:
+  1. Let |entry| be <b>[=this=]</b>'s [=FileSystemHandle/entry=].
+  1. Let |access| be the result of running |entry|'s
+     [=entry/request access=] given "`readwrite`".
+  1. If |access| is not "{{PermissionState/granted}}", reject |result| with a
+     "{{NotAllowedError}}" {{DOMException}} and abort.
```
annevk, line 276: This needs to queue a task. I guess we could postpone until #36...

index.bs
```diff
@@ -241,6 +243,66 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+
+     For files or directories with multiple hardlinks or symlinks, the entry which is deleted
+     is the entry corresponding to the path which was used to obtain the handle.
+</div>
+
+<div algorithm>
+The <dfn method for=FileSystemHandle>remove(|options|)</dfn> method steps are:
+
+1. Let |result| be [=a new promise=].
+1. Run these steps [=in parallel=]:
```
annevk, line 271: Do these steps happen in parallel, so the lock taking races with other `remove()` calls? Or should we really have some kind of "parallel queue" here? (See HTML.)

a-sully: Hmm that's a good point. I believe `removeEntry()`, `move()`, and `createSyncAccessHandle()` all have the same issue? Locking should probably not race with other operations which take a lock.

a-sully: Opened #74 for this

index.bs
a-sully: I believe this is also related to #3? Resolving this discussion in favor of having the conversation over there

index.bs
```diff
@@ -241,6 +243,66 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+
+     Attempting to delete a file or directory that does not exist is considered success,
+     while attempting to delete a non-empty directory will result in a promise rejection.
+
+  : await |handle| . {{FileSystemHandle/remove()|remove}}({ {{FileSystemRemoveOptions/recursive}}: true })
+  :: Attempts to remove the entry represented by |handle| from the underlying file system.
+
+     If the entry is a directory, its contents will also be deleted recursively.
+
+     Attempting to delete a file or directory that does not exist is considered success.
```
annevk, line 261: Is this TOCTOU or copypasta?

index.bs
```diff
@@ -241,6 +243,66 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+  1. If |access| is not "{{PermissionState/granted}}", reject |result| with a
+     "{{NotAllowedError}}" {{DOMException}} and abort.
+
+  1. Let |lockResult| be the result of [=file entry/lock/take|taking a lock=]
+     with "`exclusive`" on |entry|.
+  1. If |lockResult| is false, [=reject=] |result| with a
+     "{{NoModificationAllowedError}}" {{DOMException}} and abort.
+
+  1. If |entry| is a [=directory entry=]:
+    1. If |entry| is not an empty directory and |options|.{{FileSystemRemoveOptions/recursive}} is `false`:
```
annevk, line 284: empty directory is not a primitive we define. You can check if children is empty I suppose with the appropriate Infra references?

index.bs
```diff
@@ -241,6 +243,66 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+      1. [=/Reject=] |result| with an "{{InvalidModificationError}}" {{DOMException}} and abort.
+   1. If |entry| is the [=root directory of the origin private file system=]:
+      1. [=/Reject=] |result| with a "{{NoModificationAllowedError}}" {{DOMException}} and abort.
```
annevk, line 287: Ooh, so this is covered. It seems kinda nice to allow this, no?

a-sully: It seems useful to allow for recursive removal of all children of the root - but can we allow removal of the root itself? Could we specify that the root re-creates itself on the next call of `navigator.storage.getDirectory()`?

(outside the scope of this discussion, but if you really want to wipe the entire OPFS that seems like a use case for the Storage Buckets API)

index.bs
a-sully: copypasta. I think we should reject with `NotFoundError` if the entry does not exist? (added this in the latest patch)

index.bs
a-sully: `If |entry|'s [=directory entry/children=] is an empty set` ?

index.bs
szewai: It seems fine to allow removing root directory as it provides an easy way to clear OPFS. `navigator.storage.getDirectory()` can create the root directory again (which I think is the current behavior, i.e. we have to create the directory when clients calls `navigator.storage.getDirectory()` for the first time, or after users delete website data.)

annevk: @a-sully I suppose we could leave the root directory itself in specification-wise if that simplifies things as there's no state associated with it that would make that observable (you won't get the same object back representing it, for instance).

But yeah, as @szewai says I would expect it to act as if it was removed and the next call creates a new one.

index.bs
annevk: `If |entry|'s [=directory entry/children=] [=set/is empty=]`

index.bs
```diff
@@ -241,6 +243,67 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+  1. If |access| is not "{{PermissionState/granted}}", reject |result| with a
+     "{{NotAllowedError}}" {{DOMException}} and abort.
+
+  1. Let |lockResult| be the result of [=file entry/lock/take|taking a lock=]
+     with "`exclusive`" on |entry|.
+  1. If |lockResult| is false, [=reject=] |result| with a
+     "{{NoModificationAllowedError}}" {{DOMException}} and abort.
+
+  1. If |entry| does not exist in the underlying file system,
+     [=/reject=] |result| with a {{NotFoundError}} and abort.
```
annevk, line 281: How would you be able to take a lock if this is the case?

index.bs
a-sully: I was worried about how this interacts with https://github.com/whatwg/fs/issues/39, but I guess with directories, unlike with files, there's no way to re-create the directory once it's been removed without access to the parent (which is impossible in this case). So yeah, specifying that the next invocation of `navigator.storage.getDirectory()` creates a new OPFS seems reasonable. 

This also solves an existing problem in the Chromium implementation (mentioned in https://github.com/whatwg/fs/issues/59) where handles in the OPFS could become suddenly valid if the path is re-created from underneath it. By creating a new OPFS instance, all handles pointing to the old instance are guaranteed to never become valid again

index.bs
a-sully: It must be possible to take a lock on an entry which does not exist (for example, when [moving a file to a destination that hasn't yet been created](https://github.com/whatwg/fs/pull/10/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R306-R307))

index.bs
annevk: I suspect it won't quite solve #59 as you can still have that problem within a subdirectory or some such.

index.bs
a-sully: Ah yes, thank you

index.bs
a-sully: Correct. I was hoping we'd behave "properly" in at least this situation, but after looking into it more it turns out that the Chromium implementation always re-creates the new OPFS instance at the exact same path. So nevermind on this being an exception.

Basing a handle on a reference would solve this, and we're looking into such a re-architecture, but until that work completes I think we're unfortunately stuck with cross-browser differences on the children of deleted handles

annevk: So I think specification-wise it's probably easier if we keep the root directory around (so we don't have to meddle with the storage bucket). I think that works because you cannot observe the root directory directly (except for #59 which we have to solve anyway).

a-sully: SGTM. The latest commit adds special casing for the root

index.bs
```diff
@@ -241,6 +243,72 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+</div>
+
+<div algorithm>
+The <dfn method for=FileSystemHandle>remove(|options|)</dfn> method steps are:
+
+1. Let |result| be [=a new promise=].
+1. Run these steps [=in parallel=]:
+  1. Let |entry| be <b>[=this=]</b>'s [=FileSystemHandle/entry=].
+  1. Let |access| be the result of running |entry|'s
+     [=entry/request access=] given "`readwrite`".
```
annevk, line 271: This breaks the build currently.

annevk (commented):
Since this is a new method, should we try to do task queueing correctly from the get go?

a-sully:
> Since this is a new method, should we try to do task queueing correctly from the get go?

Do we need the new task source from #3 to be created for this, or is there some other task source we should queue to in the meantime?

a-sully:
> Since this is a new method, should we try to do task queueing correctly from the get go?

I've attempted to do this in the latest patchset (thank you for finishing off https://github.com/whatwg/storage/pull/155!), but I'm not sure whether we should be queueing everything or just the bits that may touch quota? (in this case, the actual removal of the entry). This could alternatively look like:

```
1. Let |result| be [=a new promise=].
1. Run these steps [=in parallel=]:
  1. Check permissions
  1. Grab lock (on a parallel queue)
  1. Queue a task on the storage task source to actually remove the entry
1. Return |result|.
```

...but based on your comment here (https://github.com/whatwg/fs/issues/3#issue-1070369311) it seems like the promise must be resolved/rejected on some task source, so maybe it should be this?

```
1. Let |result| be [=a new promise=].
1. Queue a global task to run these steps:
  1. Check permissions
  1. Grab lock (on a parallel queue)
  1. Queue a task on the storage task source to actually remove the entry
1. Return |result|.
```

Please let me know if you have a preference :)

index.bs
```diff
@@ -14,6 +14,8 @@ spec:webidl; type:dfn; text:resolve
 </pre>
 
 <pre class=anchors>
+urlPrefix: https://html.spec.whatwg.org/; spec: html
```
a-sully, line 17: Whoops, I meant to remove this before uploading the patch. Though if we land #94 first it won't matter anyways

annevk (commented):
This doesn't seem to have a good delineation between the main thread and the storage thread operations.

The method and the eventual task run in the main thread and can access JS.

What's missing here is an "in parallel" section that runs in some other thread and deals with locks and the file system.

Or am I misunderstanding something?

a-sully:
> This doesn't seem to have a good delineation between the main thread and the storage thread operations.
> 
> The method and the eventual task run in the main thread and can access JS.
> 
> What's missing here is an "in parallel" section that runs in some other thread and deals with locks and the file system.
> 
> Or am I misunderstanding something?

You're correct, and I've updated this in the latest patchset. It's still not perfect since file locking is still done in parallel on the main thread, but once #95 adds the new framework for file locking we can follow that pattern. I'm happy to do that in a follow-up

a-sully:
Also, https://github.com/whatwg/fs/pull/73#discussion_r1011231194 is relevant here, too... we may need to specify what taking locks on a directory looks like (i.e. you can't take the lock if a containing file is locked)

index.bs
```diff
@@ -242,6 +244,75 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+
+1. Let |result| be [=a new promise=].
+1. Run these steps [=in parallel=]:
+  1. Let |access| be the result of running [=this=]'s
+     [=FileSystemHandle/entry=]'s [=file system entry/request access=] given
+     "`readwrite`".
+  1. If |access| is not "{{PermissionState/granted}}", reject |result| with a
+     "{{NotAllowedError}}" {{DOMException}} and abort.
+
+  1. Let |entry| be [=this=]'s [=FileSystemHandle/entry=].
```
annevk, line 276: There's no this in parallel so we should move this to before going in parallel.

index.bs
```diff
@@ -242,6 +244,75 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+  1. Let |lockResult| be the result of [=file entry/lock/take|taking a lock=]
+     with "`exclusive`" on |entry|.
+  1. If |lockResult| is false, [=reject=] |result| with a
+     "{{NoModificationAllowedError}}" {{DOMException}} and abort.
+
+  1. [=Queue a task=] on the [=storage task source=] to run these steps:
```
annevk, line 282: This should use "queue a storage task"

index.bs
a-sully: Good to know. I've fixed it here, but this pattern is repeated in many other methods of this spec... to be fixed in more follow-ups!

index.bs
a-sully: Done. Please let me know if `|entry|'s [=relevant global object=]` is not correct here

index.bs
a-sully: The latest patch set is written assuming #95 will merge soon

index.bs
```diff
@@ -242,6 +244,77 @@ The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=
...
+
+1. Let |result| be [=a new promise=].
+1. Let |entry| be [=this=]'s [=FileSystemHandle/entry=].
+1. Run these steps [=in parallel=]:
+  1. Let |access| be the result of running |entry|'s
+     [=file system entry/request access=] given "`readwrite`".
+  1. If |access| is not "{{PermissionState/granted}}":
+    1. [=Queue a storage task=] given |entry|'s [=relevant global object=] to
+        [=/reject=] |result| with a "{{NotAllowedError}}" {{DOMException}}.
+    1. Abort these steps.
```
annevk, line 276: These should probably also be part of the lockResultAlgorithm? No need for nested in parallel I'd think.

a-sully: Hmm. The "request access" check should happen before attempting to take the lock, and we shouldn't bother attempting to take the lock unless the "request access" check succeeds. So I'm not sure how this would fit into the `lockResultAlgorithm`?

Should we also post the "request access" algorithm as a task? Something like:
```
(in parallel)
1. Let |lockResultAlgorithm| be an algorithm which takes a |lockResult| and
     runs these steps:
  <do file system operations>
  ...

1. Let |accessResultAlgorithm| be an algorithm which takes an |accessResult| and
     runs these steps:
  1. Reject |result| if not "granted"
  1. Take a lock on |entry| with |lockResultAlgorithm|.

1. Queue a storage task given |entry|'s relevant global object to run |entry|'s request access algorithm given "readwrite"
```

annevk: I guess you're saying that we want other lock requests to be able to happen while we do request access checks? In that case this is probably what you want.

Though what might be sufficient as I don't think this is really observable is that instead of a "file system lock queue" we have a "file system queue" or some such and all file system operations go through there.

annevk: (By the way, I don't always immediately check PRs as I expect to have to review. If you just want to hash something out https://whatwg.org/chat will likely be quicker.)

a-sully: Updated #95 to post all file system operations to a "file system queue", as we discussed offline. Once that LGTY I'll update this PR, as well. Thanks for the suggestion!

a-sully: #95 has since merged 🎉

I updated this PR to follow the new paradigm

index.bs
a-sully: This WPT has [since been updated](https://github.com/web-platform-tests/wpt/blob/2ac776fb32d8f4cfe08a41773dd4adfc6750baf6/fs/script-tests/FileSystemDirectoryHandle-removeEntry.js#L90-L103). Filed #138 to track updating `removeEntry()`

a-sully:
Hi all, I've updated this PR to make use of the various infrastructural changes that have merged recently (primarily https://github.com/whatwg/fs/pull/95 and https://github.com/whatwg/fs/pull/96). Please take a look at the updated PR. Thanks!

index.bs
```diff
@@ -416,6 +417,100 @@ The <dfn method for=FileSystemHandle>isSameEntry(|other|)</dfn> method steps are
...
+  1. Let |entry| be the result of [=locating an entry=] given |locator|.
+  1. Let |accessResult| be the result of running |entry|'s
+     [=file system entry/request access=] given "`readwrite`".
+  1. If |accessResult|'s [=file system access result/permission state=]
+     is not "{{PermissionState/granted}}", [=queue a storage task=] with
+     |global| to [=/reject=] |result| with a {{DOMException}} of
+     |accessResult|'s [=file system access result/error name=] and
+     abort these steps.
+
+  1. If |entry| is `null`, [=queue a storage task=] with |global| to [=/reject=]
```
annevk, line 461: Nit: https://infra.spec.whatwg.org doesn't use markup for null.

index.bs
```diff
@@ -416,6 +417,100 @@ The <dfn method for=FileSystemHandle>isSameEntry(|other|)</dfn> method steps are
...
+
+      1. Let |isTheRootOfABucketFileSystem| be true if
+         |isInABucketFileSystem| is true and
+         |entry|'s [=file system entry/parent=] is `null`;
+         otherwise false.
+      1. If |isTheRootOfABucketFileSystem| is true:
+        1. [=set/For each=] |child| of |entry|'s [=directory entry/children=]:
+          1. Attempt to remove |child| from the underlying file system.
+             If that throws an exception, [=/reject=] |result| with
+             that exception and abort these steps.
```
annevk, line 487: It seems weird that this would happen on the main thread.

index.bs
```diff
@@ -416,6 +417,100 @@ The <dfn method for=FileSystemHandle>isSameEntry(|other|)</dfn> method steps are
...
+       If that throws an exception, [=/reject=] |result| with that exception
+       and abort these steps.
+
+       Note: If |options|["{{FileSystemRemoveOptions/recursive}}"] is true,
+       the removal can fail non-atomically. Some files or directories might
+       have been removed while other files or directories still exist.
+
+       Issue(11): Better specify what possible exceptions this could throw.
+
+    1. [=Enqueue the following steps=] to the [=file system queue=]:
```
annevk, line 505: Do we really need nested file system queue steps?

annevk:
@fergald this too appears to have shipped in Chromium without the specification being ready with non-tentative tests?!

cc @jesup

PR #10, opened 2022-03-07
add FileSystemHandle::Move method

a-sully:
Migrated from https://github.com/WICG/file-system-access/pull/317 (see https://github.com/whatwg/fs/issues/2)

Addresses https://github.com/WICG/file-system-access/issues/64 and https://github.com/WICG/file-system-access/issues/65

See `Move.md`

   * …
   * …
   * https://github.com/web-platform-tests/wpt/blob/master/file-system-access/script-tests/FileSystemDirectoryHandle-move.js
   * Chrome: https://crbug.com/1140805
   * Firefox: …
   * Safari: …

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -148,6 +148,10 @@ interface FileSystemHandle {
   readonly attribute FileSystemHandleKind kind;
   readonly attribute USVString name;
 
+  Promise<void> move(USVString name);
+  Promise<void> move(FileSystemDirectoryHandle parent);
+  Promise<void> move(FileSystemDirectoryHandle parent, USVString name);
+
   Promise<boolean> isSameEntry(FileSystemHandle other);
 };
 </xmp>
@@ -194,6 +198,83 @@ and return {{FileSystemHandleKind/"directory"}} otherwise.
 
 The <dfn attribute for=FileSystemHandle>name</dfn> attribute must return the [=entry/name=] of the
 associated [=FileSystemHandle/entry=].
+
+### The {{FileSystemHandle/move()}} method ### {#api-filesystemhandle-move}
+
+<div class="note domintro">
+  : await |handle| . {{FileSystemHandle/move()|move}}({|new_entry_name|})
+  :: Attempts to atomically rename the entry represented by |handle| to
+     |new_entry_name| in the underlying file system.
+
+  : await |handle| . {{FileSystemHandle/move()|move}}({|destination_directory|})
+  :: Attempts to move the entry represented by |handle| to |destination_directory|,
+     while keeping its existing name.
+
+  : await |handle| . {{FileSystemHandle/move()|move}}({|destination_directory|, |new_entry_name|})
+  :: Attempts to move the entry represented by |handle| to |destination_directory|,
+     as well as renaming to |new_entry_name|.
+
+     If |destination_directory| is on the same underlying filesystem, this move
+     will be atomic. Moves to non-local file systems will are not guaranteed to
+     be atomic and will involve duplicating data.
+
+     For files or directories with multiple hardlinks or
+     symlinks, the entry which is renamed is the entry corresponding to the path
+     which was used to obtain the handle.
+
+     Attempting to move a file or directory that does not exist or passing an
+     invalid name will result in a promise rejection.
+
+</div>
+
+<div algorithm>
+The <dfn method for=FileSystemHandle>move(|destination_directory|, |new_entry_name|)</dfn> method,
+when invoked, must run these steps:
+
+1. Let |result| be [=a new promise=].
+1. Run the following steps [=in parallel=]:
+  1. Let |entry| be <b>[=this=]</b>'s [=FileSystemHandle/entry=].
+  1. Let |new_name| be |new_entry_name| if specified, otherwise [=entry/name=].
+  1. Let |new_directory| be |destination_directory| if specified, otherwise [=entry/parent=].
+  1. Let |destination| be the result of [=resolving=] |new_name|
+     relative to |new_directory|.
+  1. Let |permissionStatusSource| be the result of [=requesting access=]
+     given <b>[=this=]</b> and {{"readwrite"}}.
+     If that throws an exception, [=reject=] |result| with that exception and abort.
+  1. If |permissionStatusSource| is not {{PermissionState/"granted"}},
+     [=/reject=] |result| with a {{NotAllowedError}} and abort.
+  1. Let |permissionStatusDest| be the result of [=requesting access=]
+     given |destination| and {{"readwrite"}}.
+     If that throws an exception, [=reject=] |result| with that exception and abort.
+  1. If |permissionStatusDest| is not {{PermissionState/"granted"}},
+     [=/reject=] |result| with a {{NotAllowedError}} and abort.
+
+  1. Attempt to move |entry| to |destination| in the underlying file system.
+  1. If |entry| is a [=directory entry=]:
+    1. If |entry|'s path does not exist:
+      1. [=/Reject=] |result| with a {{NotFoundError}} and abort.
+    1. If |entry|'s path is not a directory:
+      1. [=/Reject=] |result| with a {{TypeMismatchError}} and abort.
+    1. If |entry| is in the Origin Private File System and |new_directory| is not:
+      1. [=set/For each=] |child| of |entry|'s [=directory entry/children=]:
+        1. Perform [=implementation-defined=] malware scans and safe browsing checks.
+           If these checks fail, [=/reject=] |result| with an {{AbortError}} and abort.
+    1. If the underlying move operation fails, [=/reject=] |result| with an {{AbortError}} and abort.
+  1. If |entry| is a [=file entry=]:
+    1. If |entry|'s path does not exist:
+      1. [=/Reject=] |result| with a {{NotFoundError}} and abort.
+    1. If |entry|'s path is not a file:
+      1. [=/Reject=] |result| with a {{TypeMismatchError}} and abort.
+    1. If |entry| is in the Origin Private File System and |destination_directory| is not:
+      1. Perform [=implementation-defined=] malware scans and safe browsing checks.
+         If these checks fail, [=/reject=] |result| with an {{AbortError}} and abort.
+    1. If the underlying move operation fails, [=/reject=] |result| with an {{AbortError}} and abort.
+
+  1. [=/Resolve=] |result| with `undefined`.
+
+1. Return |result|.
+
+</div>
 
 ### The {{FileSystemHandle/isSameEntry()}} method ### {#api-filesystemhandle-issameentry}
 
--- a/Move.md
+++ b/Move.md
@@ -1,0 +1,148 @@
+# Adding the FileSystemHandle::move() method
+
+## Authors:
+
+* Austin Sullivan (asully@chromium.org)
+
+## Participate
+
+* [Issue tracker](https://github.com/whatwg/fs/issues)
+
+## Table of Contents
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Adding the FileSystemHandle::move() method](#adding-the-filesystemhandlemove-method)
+  - [Authors:](#authors)
+  - [Participate](#participate)
+  - [Table of Contents](#table-of-contents)
+- [Overview](#overview)
+- [API Surface](#api-surface)
+- [Handling Moves to Other Directories](#handling-moves-to-other-directories)
+  - [(1) Moving to a Non-Local File System](#1-moving-to-a-non-local-file-system)
+  - [(2) Moving out of the OPFS](#2-moving-out-of-the-opfs)
+  - [(3) Moving to a Directory on the Local File System](#3-moving-to-a-directory-on-the-local-file-system)
+  - [Open Questions](#open-questions)
+    - [What, if any, checks need to be run on same-file-system moves?](#what-if-any-checks-need-to-be-run-on-same-file-system-moves)
+    - [What should happen if the operation cannot be completed successfully?](#what-should-happen-if-the-operation-cannot-be-completed-successfully)
+- [Implementation Notes](#implementation-notes)
+- [Alternatives Considered](#alternatives-considered)
+  - [Only support "rename" functionality](#only-support-rename-functionality)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Overview
+
+Currently, the API does not support an efficient way to move or rename files or
+directories. This requires creating a new file/directory, copying over data
+(recursively, in the case of a directory), and removing the original. This
+process is slow, error prone (e.g. disk full), and can require re-uploading
+files in some cases (e.g. Google Drive files on Chrome OS).
+
+We propose adding a new `FileSystemHandle::move()` method. `move()` will offer
+atomic moves of a file or directory if it is moved within the same file system,
+while moves to non-local file systems will not be guaranteed to be atomic and
+may involve duplicating data. See the [Open Questions](#open-questions).
+
+# API Surface
+We propose adding an overloaded `move()` method with three variations.
+
+```
+interface FileSystemHandle {
+    ...
+    Promise<void> move(USVString new_entry_name);
+    Promise<void> move(FileSystemDirectoryHandle destination_directory);
+    Promise<void> move(FileSystemDirectoryHandle destination_directory, USVString new_entry_name);
+    ...
+};
+```
+
+`move(new_entry_name)` will rename a file or directory to `new_entry_name`. It
+is guaranteed to be atomic and will not involve duplicating data.
+
+`move(destination_directory)` allows a user to specify a `destination_directory`
+to move to, which may or may not be on the same file system. The name of the
+original file or directory is retained. This is not guaranteed to be atomic,
+since the destination directory may be on a different file system.
+
+`move(destination_directory, new_entry_name)` allows a user to rename an entry
+and while moving it to a new directory.
+
+# Handling Moves to Other Directories
+
+There are three cases to consider when moving to other directories:
+
+## (1) Moving to a Non-Local File System
+
+Moving across file systems cannot be guaranteed to be atomic if the operation
+cannot be completed successfully (e.g. network connection issues, lack disk
+space, etc.). This may result in partial writes to the target directory.
+
+Note that this approach is no worse than what is available currently.
+
+## (2) Moving out of the OPFS
+
+[#310](https://github.com/WICG/file-system-access/pull/310) exposes a strong use
+case for fast, atomic moves to other directories on the same file system. Files
+can be written efficiently in the Origin Private File System using an
+Access Handle, then moved to a directory of the user's choice. However, writes
+using an AccessHandle are not subjected to Safe Browsing checks in Chrome.
+Moving files out of the OPFS to a user's directory will require running Safe
+Browsing checks and adding the mark-of-the-web on all moved files.
+
+For example, moving a large directory across the OPFS -> file system boundary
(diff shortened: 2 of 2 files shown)


annevk:
I'd suggest we don't include the `Move.md` file here.

Explosion-Scratch:
Any updates on this? Thanks @a-sully for working on this though. When this gets merged what exactly happens? Does this feature get added to all JavaScript engines supporting the whatwg standard?

jesup:
Open questions:

- What happens if there are existing SyncAccessHandles to a file being moved?
- What if there are FileHandles to a file being moved?
- What if there are DirectoryHandles to a file being moved?
- What happens if a different process/worker/etc tries to grab a FileHandle/DirectoryHandle/SyncAccess handle on a file or directory being moved while it is still in the process of being moved (promise has not resolved yet).   If this does not succeed, what error is thrown?
- Is it possible that a move() could end up in a state other than completely moved or completely not-moved?   I.e. if there's a disk or permission error during moving a directory.   This is unlikely in OPFS, but may be possible in the more general usage.

jesup:
> Open questions:
> 
>     * What happens if there are existing SyncAccessHandles to a file being moved?
> 
>     * What if there are FileHandles to a file being moved?
> 
>     * What if there are DirectoryHandles to a file being moved?
> 
This comment in move.md from the patch partly answers these questions (I presume what's meant here is "All handles to entries"):
- All entries contained within a directory which is moved no longer point to an
  existing file/directory. We currently have no plans to explicitly update or
  invalidate these handles, which would otherwise be an expensive recursive
  operation.

That doesn't answer about FileHandles or SyncAccessHandles to a file being moved/renamed.

>     * What happens if a different process/worker/etc tries to grab a FileHandle/DirectoryHandle/SyncAccess handle on a file or directory being moved while it is still in the process of being moved (promise has not resolved yet).   If this does not succeed, what error is thrown?
> 
>     * Is it possible that a move() could end up in a state other than completely moved or completely not-moved?   I.e. if there's a disk or permission error during moving a directory.   This is unlikely in OPFS, but may be possible in the more general usage.

This is discussed in move.md, but there's no resolution of the discussion.

jesup:
Right now the justification for the answer about moving directories is based on move.md, which isn't here.
There's also no description in the spec as to what happens to any of these cases; these need to be speced.    One can infer that if we move foo to bar, that the directory now holds bar instead of foo.   But there may be race conditions between a worker and mainthread, or worker and worker...
It also means that one needs to reverse-engineer the current implementations.

jesup:
Note that the WPT tests for move() encode several things not in this spec related to what are valid names:

testing that you can't rename a file to "Lorem." (no trailing periods - this isn't specified)
testing that you can't rename a file "#$23423@352^*3243" (no invalid characters - this isn't specified - none of those are separators)

The spec says: "A valid file name is a [string](https://infra.spec.whatwg.org/#string) that is not an empty string, is not equal to "." or "..", and does not contain '/' or any other character used as path separator on the underlying platform." 
and
"Note: This means that '\' is not allowed in names on Windows, but might be allowed on other operating systems. Additionally underlying file systems might have further restrictions on what names are or aren’t allowed, so a string merely being a [valid file name](https://whatpr.org/fs/21.html#valid-file-name) is not a guarantee that creating a file or directory with that name will succeed." "We should consider having further normative restrictions on file names that will never be allowed using this API, rather than leaving it entirely up to underlying file systems."

jesup:
The tests also test that move() can overwrite an existing file.  This is mentioned nowhere in the spec, and the spec says "Attempt to move entry to destination in the underlying file system." which does not to me indicate it should overwrite.  (It also checks that having a WritableFileStream open on the destination blocks move())

jesup:
Also file.move(dir, "") should fail, per the spec; the wpt tests assume it's the same as file.move(dir)

a-sully:
Thanks for bringing up these questions! Apologies for the confusion, Move.md is a bit outdated since it was written before the locking mechanisms for SyncAccessHandles were created. I'm planning to update this PR once #21 lands to incorporate those locking mechanisms. For now, to respond to your questions...

> Open questions:
> 
> * What happens if there are existing SyncAccessHandles to a file being moved?

`move()` takes an exclusive lock on the handle. Therefore a file cannot be moved while there is an open SyncAccessHandle (a reminder to me that we really need to clear up #18)

> * What if there are FileHandles to a file being moved?
> * What if there are DirectoryHandles to a file being moved?

#30 is intended to address these. Since a FileSystemHandle maps to a path, these handles will dangle once the underlying file/dir is moved.

> * What happens if a different process/worker/etc tries to grab a FileHandle/DirectoryHandle/SyncAccess handle on a file or directory being moved while it is still in the process of being moved (promise has not resolved yet).   If this does not succeed, what error is thrown?

If a SyncAccessHandle is being created, presumably whichever operation is able to grab their respective lock will win and the other will reject.

As for `getFileHandle()` and `getDirectoryHandle()`, if the path has been moved (and `create: true` has not been specified) then I don't think this deserves any special treatment? Do you see a problem with rejecting with `NotFoundError` as is currently specified when the requested path does not exist?

> * Is it possible that a move() could end up in a state other than completely moved or completely not-moved?   I.e. if there's a disk or permission error during moving a directory.   This is unlikely in OPFS, but may be possible in the more general usage.

Yes. A couple examples:
- Power loss during a cross-file system move
- Moving a directory from OPFS -> non-OPFS will trigger Safe Browsing checks on all containing files. If any of these checks fail, the unsafe files will not be moved (while the safe files will be moved)

In these cases, the promise will be rejected with an `AbortError`.

> The tests also test that move() can overwrite an existing file. This is mentioned nowhere in the spec, and the spec says "Attempt to move entry to destination in the underlying file system." which does not to me indicate it should overwrite. (It also checks that having a WritableFileStream open on the destination blocks move())

You are correct. And I'm wondering whether this is the right behavior in the first place... I think there's a strong argument to be made that we should reject a move to a path that already exists and allow the site to delete the destination handle and try again if they truly want it overwritten.

a-sully:
> Also file.move(dir, "") should fail, per the spec; the wpt tests assume it's the same as file.move(dir)

Good catch. We should probably update the WPTs and not the spec? Ideally the "is this a safe file" code can be shared between `move()`,  `getFileHandle()`, and `getDirectoryHandle()`

jesup:
> > Also file.move(dir, "") should fail, per the spec; the wpt tests assume it's the same as file.move(dir)
> 
> Good catch. We should probably update the WPTs and not the spec? Ideally the "is this a safe file" code can be shared between `move()`, `getFileHandle()`, and `getDirectoryHandle()`

yes, the WPTs should be updated

jesup:
Attempts to move a directory to within itself should fail.
Attempts to move the root directory should fail... at least in OPFS, but that would be handled by the above restriction anyways.   No idea about non-OPFS.
Note that for non-OPFS if you move a directory out of OPFS you have to check all files in the directory for safe-browsing

jesup:
The wpt tests check that moving a file to the same name succeeds (I.e. succeeds in doing nothing).  (They don't check if moving a directory to the same name succeeds, note.)   This behavior does not appear to be anywhere in the spec here; it basically says "attempts to move it in the underlying filesystem" which could fail or succeed.
The operation here should be well-specified, probably by saying that it should always succeed.

jesup:
I wonder if the fact that move-that-doesn't-move succeeds in chrome is related to the fact that it currently silently allows a move() to overwrite a file.   At this moment, we fail that WPT test because we check if the destination file exists and if so, fail.

index.bs
```diff
@@ -234,6 +238,110 @@ The <dfn attribute for=FileSystemHandle>kind</dfn> getter steps are to return
...
+         If these checks fail, [=/reject=] |result| with an
+         "{{AbortError}}" {{DOMException}} and abort.
+
+  1. Attempt to move |entry| to |destination_entry| in the underlying file system.
+    1. If the underlying move operation throws an exception, [=/reject=]
+       |result| with that exception and abort.
+
+     Issue(68): Better specify what possible exceptions this could throw.
+
+  1. Set |entry|'s [=entry/name=] to |new_name|.
```
a-sully, line 335: note: we should also specify that |entry|'s parent is now |new_directory|. Not sure how this should be worded until we have resolution on #59

index.bs
```diff
@@ -188,6 +188,10 @@ interface FileSystemHandle {
   readonly attribute FileSystemHandleKind kind;
   readonly attribute USVString name;
 
+  Promise<undefined> move(USVString new_entry_name);
+  Promise<undefined> move(FileSystemDirectoryHandle destination_directory);
+  Promise<undefined> move(FileSystemDirectoryHandle destination_directory, USVString new_entry_name);
```
annevk, line 193: Hmm, this results in somewhat convoluted overloading I thought we were trying to move away from. @domenic what do you think? Should this use a dictionary?

annevk: Also generally we don't follow Python and instead use camelCase for variable names.

domenic: I think the overloading might be OK here, although you could reduce it by combining the latter two into `move(FileSystemDirectoryHandle destinationDirectory, optional USVString newEntryName)`.

In terms of API design, I think all of the following are worth considering:

- Positional parameters, varying order (this proposal):
  - `fileHandle.move("renamed.txt")`
  - `fileHandle.move(directoryHandle)`
  - `fileHandle.move(directoryHandle, "renamed.txt")`
- Positional parameters, consistent order (this will choke on https://github.com/plinss/widlparser/issues/79 but is still valid...)
  - `fileHandle.move("renamed.txt")`
  - `fileHandle.move(undefined, directoryHandle)`
  - `fileHandle.move("renamed.txt" directoryHandle)`
- Directory in an options bag (still requires overloading)
  - `fileHandle.move("renamed.txt")`
  - `fileHandle.move({ directory: directoryHandle })`
  - `fileHandle.move("renamed.txt", { directory: directoryHandle })`
- Options bag for everything
  - `fileHandle.move({ name: "renamed.txt"} )`
  - `fileHandle.move({ directory: directoryHandle })`
  - `fileHandle.move({ name: "renamed.txt", directory: directoryHandle })`
- Separate out move within directory vs. move to a new directory
  - `fileHandle.move("renamed.txt")` (or maybe `fileHandle.rename("renamed.txt")`)
  - `fileHandle.moveTo(directoryHandle)`
  - `fileHandle.moveTo(directoryHandle, "renamed.txt")`
- Path-based version?? Seems counter to how much of the rest of the API works, but is more what I would expect from other FS APIs.
  - `fileHandle.move("renamed.txt")`
  - `fileHandle.move(directoryHandle.resolve(fileHandle.name))` (not great...)
  - `fileHandle.move(directoryHandle.resolve("renamed.txt"))`

On balance I think all of these are pretty reasonable, including the current proposal.

Apologies if this has been discussed before; I didn't see it in a quick skim of the several issues and PRs that seem to be about this feature.

a-sully: On Chromium we already support `move()` for files within the OPFS, so if there's not a strong preference towards any one option I'd prefer to not make any breaking changes. Combining the latter two to make `newEntryName` optional seems reasonable, though, and wouldn't be a breaking change.

If we're convinced we need to change this, the options bag would be my preference since it's more extensible if we want to add more options in the future (e.g. "should this overwrite an existing file at this location?").

Regarding the `rename()` and `moveTo` proposals, I had previously proposed a separate `rename()` method for the first case, (i.e. `fileHandle.rename("renamed.txt")`) but received feedback from developers that a consolidated `move()` method was preferred.

I would especially recommend against a path-based version, since I think it could be confusing to only support relative paths and absolute paths don't make sense outside of the OPFS and won't make sense once sites have access to multiple OPFS instances when Storage Buckets is eventually supported. I'd much rather make it easier to get the destination directory through a convenience method like the one proposed here https://github.com/WICG/file-system-access/issues/378

a-sully: It's not the most pretty, but it seems reasonable enough to me

a-sully:
We've received feedback from partners that we should not reject moves that would otherwise overwrite a file. Emscripten (@tlively) mimics the posix interface, which expects `move()` to be able to overwrite a file. The API doesn't afford a safe solution for emulating overwriting moves, and it's not great to require application code to change to comply with this.

We could introduce this as an opt-in (as I suggested earlier https://github.com/whatwg/fs/pull/10#discussion_r1018585899) if we think the default should be not to allow overwrites. Thinking about the non-OPFS use cases, at the very least I'd like to advocate that user agents should be able to reject a rename operation if the site does not have access to the parent directory

annevk:
> I'd like to advocate that user agents should be able to reject a rename operation if the site does not have access to the parent directory.

That seems more like an access failure and not related to move semantics.

Do we have use cases where you want to throw for overwrites? Can they be solved through some other way?

a-sully:
> That seems more like an access failure and not related to move semantics.

Fair enough

> Do we have use cases where you want to throw for overwrites? Can they be solved through some other way?

My expectation is that C code being ported to the web via WASM will expect to be able to overwrite (as posix allows), but for web-native applications (especially those outside the OPFS), not allowing overwriting seems like the safer thing to do, since accidental overwrites would result in data loss. 

In both cases, you can _somewhat_ mimic the opposite behavior. If you have write access to the destination directory and the default is...
- allow overwrites, the site can choose to `removeEntry()` on the entry to be overwritten before moving
  - If for some reason `removeEntry()` succeeds and `move()` fails, this will result in data loss
- reject overwrites, the site can call `getFileHandle(name, { create: false })` or `getDirectoryHandle(name, { create: false })` on the entry to be overwritten to see if it exists before moving
  - Technically vulnerable to TOCTOU errors

annevk:
TIL about TOCTOU, makes sense. So I guess being able to pass some option to `move()` along the lines of `throwOnReplace` would help, so it's all atomic.

I just noticed that @jesup wrote above that Firefox currently throws for overwrites. Is matching POSIX here not as accepted as I would have assumed?

annevk:
I chatted about this with @szewai and she made the point that applications that are concerned about TOCTOU can use Web Locks. And that yeah, POSIX is expected.

a-sully:
Fair enough. I can update the WPTs and spec to assert that overwrites are allowed

jesup:
From https://man7.org/linux/man-pages/man2/rename.2.html POSIX rename() atomically replaces 'newfile' on rename() if it exists, so you're correct that posix wants that behavior. 
I think at least some non-normative mention of WebLocks to avoid overwriting would be helpful here.   While I might prefer 2 to 3, I don't think that's a big deal.   I probably prefer 1 to either 2 or 3, but I don't think there's a requirement to go that way; I just think the API would be better that way; the same functionality is available with all 3 options (more or less).   (Just another area that shows it's better to get agreement on APIs first, before any field lockin occurs... like with SyncAccessHandle synchronous-ness)

a-sully:
> Just another area that shows it's better to get agreement on APIs first, before any field lockin occurs... like with SyncAccessHandle synchronous-ness

Agreed!

On that note, we're planning to ship `move()` of files outside of the OPFS on Chromium browsers soon. Please feel free to leave feedback on the explainer (https://github.com/a-sully/fs/pull/2) or TAG review (https://github.com/w3ctag/design-reviews/issues/805)

Q: Should we scope down this PR to only consider file moves?  Ideally this PR would have merged months (when we decided to support `move()` for files within the OPFS) but I worry it will be tricky to land this PR while it contains directory moves while #59 is unresolved. Thoughts?

index.bs
```diff
@@ -416,6 +419,211 @@ The <dfn method for=FileSystemHandle>isSameEntry(|other|)</dfn> method steps are
...
+  1. Let |isTheRootOfABucketFileSystem| be true if
+     |isInABucketFileSystem| is true and
+     |entry|'s [=file system entry/parent=] is `null`;
+     otherwise false.
+  1. If |isTheRootOfABucketFileSystem| is true,
+     [=queue a storage task=] with |global| to [=/reject=] |result| with a
+     "{{InvalidModificationError}}" {{DOMException}} and abort these steps.
+
+  Issue: Do not allow |entry| to be moved if it is a
+  <a href=https://wicg.github.io/file-system-access/#wellknowndirectory-too-sensitive-or-dangerous>sensitive directory</a>
```
a-sully, line 552: Whoops, I meant to say that [well-known directories](https://wicg.github.io/file-system-access/#enumdef-wellknowndirectory) should not be moved. The site should never obtain a handle to a sensitive directory in the first place

a-sully:
Hi all, I've updated this PR to make use of the various infrastructural changes that have merged recently (primarily #95 and #96). Please take a look at the updated PR. Thanks!

@jesup I've tried to answer all your questions. Apologies for not being able to answer them months ago! (We really needed the above infrastructure changes to appropriately answer many of them.) Please let me know if there's anything I've missed. 

> Open questions:
> 
> * What happens if there are existing SyncAccessHandles to a file being moved?

`move()` takes a lock on both the source and destination file. If there is an open SyncAccessHandle, neither the file itself nor any of its parents (as per https://github.com/whatwg/fs/pull/139#pullrequestreview-1528941955) are able to be moved

> * What if there are FileHandles to a file being moved?

Only the handle which `move()` is called on is updated. I've added a note about this in the PR

> * What if there are DirectoryHandles to a file being moved?

Ditto

> * What happens if a different process/worker/etc tries to grab a FileHandle/DirectoryHandle/SyncAccess handle on a file or directory being moved while it is still in the process of being moved (promise has not resolved yet).   If this does not succeed, what error is thrown?

We now have a [file system queue](https://fs.spec.whatwg.org/#file-system-queue)!

> * Is it possible that a move() could end up in a state other than completely moved or completely not-moved?   I.e. if there's a disk or permission error during moving a directory.   This is unlikely in OPFS, but may be possible in the more general usage.

Yes. I've added a note about this in the PR, though since no browser has yet implemented directory moves outside of the BucketFS, that part is admittedly a bit vague for now

> * I think at least some non-normative mention of WebLocks to avoid overwriting would be helpful here
> * Attempts to move a directory to within itself should fail.
> * Attempts to move the root directory should fail... at least in OPFS

Added!

> Note that for non-OPFS if you move a directory out of OPFS you have to check all files in the directory for safe-browsing

As per [this](https://github.com/whatwg/fs/blob/main/proposals/MovingNonOpfsFiles.md#what-about-moving-files-from-the-opfs-to-user-visible-directories-or-vice-versa), I've specified that moving across the OPFS boundary is not allowed (for now). I welcome thoughts on #114, as well as thoughts on supporting cross-BucketFS moves

> * The wpt tests check that moving a file to the same name succeeds (I.e. succeeds in doing nothing).

I agree that we shouldn't need to actually call into e.g. `mv`, but I don't think we should do _nothing_. We should definitely run permission checks, for example, and it seems useful to still e.g. get a `NoModificationAllowedError` if the file is locked. For now I've specified it such that we'll run most all the steps right up until `mv`, but exit early before actually attempting to move on disk.

a-sully:
A few things to note:
- The steps for requesting access to the destination are still subject to change depending on what we decide in #131
- Locking a directory entry is not yet supported, but we're working on it: #139
- No browsers yet support directory moves outside of the BucketFS, so there are likely more restrictions we'll want to add - that may require some acrobatics with the https://github.com/wicg/file-system-access spec. For example, we should not allow [well-known directories](https://wicg.github.io/file-system-access/#enumdef-wellknowndirectory) to be moved

JasonGrass:
The behavior of renaming files is inconsistent with the native API.

When using the `FileSystemHandle.move` method to rename files, it modifies the "Modified time" of the file. However, this issue does not occur when manually operating or using the native API, as renaming usually does not modify a file's "Modified time".
Is this behavior expected?

Should I report this issue here or somewhere else?

PS My test environment: Win10, Chrome120.0.6099.129

image

miketaylr:
> Is this behavior expected?

AFAICT, this PR doesn't say anything about setting a file's `lastModified` (or to not do so). @a-sully?

jozefchutka:
Not sure why this remains open but the data seems also incomplete on https://github.com/mdn/browser-compat-data . I have created https://github.com/mdn/browser-compat-data/issues/26876 to update the support list as it seems chrome, firefox and safari all support this feature

PR #21, opened 2022-04-11
Add Access Handles to spec

fivedots:
This PR adds Access Handles to the spec. The Access Handles surface differs from existing ones by offering in-place and exclusive write access to a file’s content. This change, along with the ability to consistently read unflushed modifications and the availability of a synchronous variant on dedicated workers, significantly improves performance and unblocks new use cases.

There are still a few open TODOs where advice would be appreciated! 

The PR also updates AccessHandles.md to reflect changes in https://github.com/WICG/file-system-access/pull/367.

   * Chrome is in the process of shipping
   * Safari has shipped
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://github.com/web-platform-tests/wpt/tree/master/file-system-access
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1218431
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1761370
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=231185

***
<a href="https://whatpr.org/fs/21.html" title="Last updated on Sep 29, 2022, 5:27 PM UTC (c7b7457)">Preview</a> | <a href="https://whatpr.org/fs/21/464b505...c7b7457.html" title="Last updated on Sep 29, 2022, 5:27 PM UTC (c7b7457)">Diff</a>


fivedots:
A first pass review was done by @domenic in https://github.com/WICG/file-system-access/pull/344.

tomayac (commented):
```js
// In all contexts
const accessHandle = await fileHandle.createAccessHandle();
await accessHandle.writable.getWriter().write(buffer);
const reader = accessHandle.readable.getReader({ mode: "byob" });
// Assumes seekable streams, and SharedArrayBuffer support are available
await reader.read(buffer, { at: 1 });

// Only in a worker context
const accessHandle = await fileHandle.createSyncAccessHandle();
const writtenBytes = accessHandle.write(buffer);
const readBytes = accessHandle.read(buffer, { at: 1 });
```

```js
const accessHandle1 = await fileHandle.createAccessHandle();
try {
  const accessHandle2 = await fileHandle.createAccessHandle();
} catch (e) {
  // This catch will always be executed, since there is an open access handle
}
await accessHandle1.close();
// Now a new access handle may be created
```

fivedots:
Updated examples on AccessHandles.md as proposed by @tomayac.

annevk:
I don't think we should keep AccessHandle.md as a separate file. If the information in there is relevant to the standard, it should be folded into the main document. If it's not, it should probably be maintained elsewhere. At least we don't usually have such documents alongside WHATWG standards.

index.bs
```diff
@@ -87,8 +87,40 @@ never be allowed using this API, rather than leaving it entirely up to underlyin
...
+1. If |value| is "`exclusive`":
+   1. If |lock| is "`open`":
+     1. Set lock to "`taken-exclusive`".
+     1. Return `true`.
+1. If |value| is "`shared`":
+   1. If |lock| is "`open`":
+     1. Set lock to "`taken-shared`".
+     1. Return `true`.
+   1. Else if |lock| is "`taken-shared`":
+     1. Return `true`.
```
annevk, line 108:
```suggestion
   1. Otherwise, if |lock| is "`taken-shared`", return `true`.
```
(We don't use "Else", this also applies elsewhere.)

index.bs
```diff
@@ -87,8 +87,40 @@ never be allowed using this API, rather than leaving it entirely up to underlyin
...
+
+<div algorithm>
+To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of "`exclusive`" or "`shared`" on a given [=file entry=] |file|,
+run the following steps:
+
+1. Let |lock| be the |file|'s [=file entry/lock=].
+1. If |value| is "`exclusive`":
+   1. If |lock| is "`open`":
+     1. Set lock to "`taken-exclusive`".
+     1. Return `true`.
```
annevk, line 102: Style nit: we don't use markup for boolean values, see https://infra.spec.whatwg.org/#booleans.

index.bs
```diff
@@ -314,6 +354,49 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method,
 
 </div>
 
+### The {{FileSystemFileHandle/createSyncAccessHandle()}} method ### {#api-filesystemfilehandle-createsyncaccesshandle}
+
+<div class="note domintro">
+  : |handle| = await |fileHandle| . {{FileSystemFileHandle/createSyncAccessHandle()|createSyncAccessHandle}}()
+  :: Returns a {{FileSystemSyncAccessHandle}} that can be used to read/write from/to the file.
```
annevk, line 361:
```suggestion
  :: Returns a {{FileSystemSyncAccessHandle}} that can be used to read from/write to the file.
```

index.bs
```diff
@@ -314,6 +354,49 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method,
...
+     Changes made through |handle| might be immediately reflected in the file represented by |fileHandle|.
+     To ensure the changes are reflected in this file, the handle must be flushed or closed.
```
annevk, line 363: You cannot use normative terms such as "must" in non-normative text. See https://infra.spec.whatwg.org/#conformance for suggestions.

index.bs
```diff
@@ -314,6 +354,49 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method,
...
+
+1. Let |result| be [=a new promise=].
+1. Run the following steps [=in parallel=]:
+  1. Let |access| be the result of running [=this=]'s [=FileSystemHandle/entry=]'s
+     [=entry/request access=] given "`readwrite`".
+     If that throws an exception, [=reject=] |result| with that exception and abort.
+  1. If |access| is not "{{PermissionState/granted}}",
+     reject |result| with a {{NotAllowedError}} and abort.
+  1. Let |entry| be [=this=]'s [=FileSystemHandle/entry=].
+  1. If |entry| does not represent an [=/entry=] in an [=origin private file system=],
```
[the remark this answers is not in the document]
annevk: Does using "NotAllowedError" actually solve the problem? The first guard reuses the exception from "request access" (if any) and it seems like that could in theory throw arbitrary exceptions.

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+
+</xmp>
+
+A {{FileSystemSyncAccessHandle}} has an associated <dfn for=FileSystemSyncAccessHandle>\[[file]]</dfn>
+(a [=file entry=]).
+
+A {{FileSystemSyncAccessHandle}} has an associated <dfn for=FileSystemSyncAccessHandle>\[[state]]</dfn>,
+a string that may exclusively be "`open`" or "`closed`".
+
+<div class="note domintro">
```
annevk, line 1051: This is not a domintro.

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+
+A {{FileSystemSyncAccessHandle}} is an object that is capable of reading/writing from/to,
```
annevk, line 1053:
```suggestion
A {{FileSystemSyncAccessHandle}} is an object that is capable of reading from/writing to,
```

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+1. Let |fileSize| be |fileContents|'s [=byte sequence/length=].
+1. Let |readStart| be |options|.{{FileSystemReadWriteOptions/at}}.
+1. If |readStart| is larger than |fileSize|, return 0.
+1. Let |readEnd| be |readStart| + (|bufferSize| &minus; 1).
+1. If |readEnd| is larger than |fileSize|, set |readEnd| to |fileSize|.
+1. Let |result| be |bufferSize|.
+1. Let |bytes| be a [=byte sequence=] containing the bytes from |readStart| to |readEnd| of |fileContents|.
+1. If the operations reading from |fileContents| in the previous steps failed:
+    1. If there were partial reads  and the number of modified bytes in |bytes|  is known,
+       set |result| to the number of modified bytes.
```
annevk, line 1097: Why would bytes be modified during a read?

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+    1. Else set |result| to 0.
+1. Let |arrayBuffer| be |buffer|'s [=underlying buffer=].
+1. [=write|Write=] |bytes| into |arrayBuffer|.
+1. Return |result|.
+
+</div>
+
+### The {{FileSystemSyncAccessHandle/write()}} method ### {#api-filesystemsyncaccesshandle-write}
+
+<div class="note domintro">
```
annevk, line 1107: I think this only works with `<dl class=domintro>`...

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+1. Let |data| be [=get a copy of the buffer source|a copy of=] |buffer|.
+
+   Note: The objective of this copy is to prevent concurrent modifications of the buffer
+   from being reflected in the file. Depending on the implementation, a copy
+   of non-shared ArrayBuffers may not be needed in a synchronous call like this one.
+   It might also be possible to do a buffer copy at an earlier time or through a
+   different mechanism.
+
+1. If |writePosition| is larger than |oldSize|,
+   append |writePosition| &minus; |oldSize| `0x00` (NUL) bytes to the end of |fileContents|.
```
annevk, line 1132: Nit: bytes don't have markup: https://infra.spec.whatwg.org/#byte.

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+<div algorithm>
+The <dfn method for=FileSystemSyncAccessHandle>getSize()</dfn> method, when invoked, must run
+these steps:
+
+1. If [=this=].[=[[state]]=] is "`closed`", return [=a promise rejected with=] an {{InvalidStateError}}.
+1. Let |p| be [=a new promise=] created in the [=relevant Realm=] of [=this=].
+1. Run the following steps [=in parallel=]:
+  1. Let |size| be the [=byte sequence/length=] of [=this=].[=[[file]]=]'s [=file entry/binary data=].
+  1. [=/Resolve=] |p| with |size|.
+1. Return |p|.
```
annevk, line 1206: It seems a little weird that `read()` and `write()` are synchronous and obtain the size information as well, but here the length is behind a promise. I guess that also applies to `truncate()`. Could you explain how that makes sense?

annevk (commented):
I took a pass at this and mainly found style nits. Hope that helps. Thanks for pushing this forward!

jesup:
I notice that while the webidl indicates that SyncAccessHandle is for workers only, the commentary for webdevelopers doesn't mention workers anywhere.

I also note that there's no definition of AccessHandles (such as you see in https://github.com/WICG/file-system-access/blob/main/AccessHandle.md#exposing-accesshandles-on-all-filesystems), which means the only way to read data is on worker via SyncAccessHandles (WritableFileStreams only support writing data here).   I presume this is an intentional decision made sometime since AcessHandle.md was written?    And this means there's no way to read data in an async fashion, it's sync only.

There's also no mention of OPFS until the very end of the document; it's referenced by description in the intro, but not by name.

And ditto to anne's points, especially as to why truncate/getSize are async (since they're only for SyncAccessHandle).

index.bs
```diff
@@ -87,8 +87,40 @@ never be allowed using this API, rather than leaving it entirely up to underlyin
...
+1. Return `false`.
+
+</div>
+
+<div algorithm>
+To <dfn for="file entry/lock">release</dfn> a [=file entry/lock=] on a given [=file entry=] |file|,
+run the following steps:
+
+1. Let |lock| be the |file|'s associated [=file entry/lock=].
+1. Set |lock| to "`open`".
```
jesup, line 118: This says that if 10 WritableFileStreams are created with shared locks, and one of the WritableFileStreams goes away, it will cause the lock to be dropped.

Though I'll also note that it's unclear *how* a WritableFileStream's being destroyed would lead to a 'release' occurring -- unlike SyncAccessHandles, there's no 'close', so it would be dependent on garbage collection, and we generally want GC to not be visible to applications.  Alternatively, this may mean that once a WritableFileStream is created for a file, you can never get a SyncAccessHandle to that file (unless the browser is restarted, or perhaps after all instances of that origin have gone away).
In any case, this is confusing and needs to be much more clearly specified.

jesup: Also odd with WritableFileStreams: they support (only) write(), seek() and truncate() (and they keep the "writes to a backup of the file).   No way to read the data (except with an SyncAccessHandle).  And no way to read an OPFS file except on a Worker (since the non-sync AccessHandle seems to have disappeared).

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+
+1. If [=this=].[=[[state]]=] is "`closed`", throw an {{InvalidStateError}}.
+1. Let |bufferSize| be |buffer|'s [=byte length=].
+1. Let |fileContents| be [=this=].[=[[file]]=]'s [=file entry/binary data=].
+1. Let |fileSize| be |fileContents|'s [=byte sequence/length=].
+1. Let |readStart| be |options|.{{FileSystemReadWriteOptions/at}}.
+1. If |readStart| is larger than |fileSize|, return 0.
+1. Let |readEnd| be |readStart| + (|bufferSize| &minus; 1).
+1. If |readEnd| is larger than |fileSize|, set |readEnd| to |fileSize|.
+1. Let |result| be |bufferSize|.
```
jesup, line 1093: This probably though be:
Let result be (readEnd - readStart) + 1 (since we can change readEnd right above).

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+
+### The {{FileSystemSyncAccessHandle/write()}} method ### {#api-filesystemsyncaccesshandle-write}
+
+<div class="note domintro">
+  : |handle| . {{FileSystemSyncAccessHandle/write()|write}}(|buffer|)
+  : |handle| . {{FileSystemSyncAccessHandle/write()|write}}(|buffer|, { {{FileSystemReadWriteOptions/at}} })
+  :: Writes the content of |buffer| into the file associated with |handle|, optionally at a given offset.
+</div>
+
+// TODO(fivedots): Figure out how to properly check the available storage quota (in this method and others) by passing the right storage shelf.
```
jesup, line 1113: Any plans to resolve this todo in a manner that we can implement (without all of Storage API)?   What does chrome's current implementation do, and what does Safari do?

a-sully (commented):
Apologies for not looking at this sooner! I've added some notes about Chrome's current implementation, which I hope can serve as a starting point but which of course we can readily change

jesup:
> See [this doc](http://doc/1lsQhTsfcVIeOW80dr467Auud_VCeAUv2ZOkC63oSyKo) for the justification behind the split sync/async interface.

I think you mean [this](https://docs.google.com/document/d/1lsQhTsfcVIeOW80dr467Auud_VCeAUv2ZOkC63oSyKo/edit?pli=1#)

index.bs
jesup: The real missing part here is specifying what happens in quota cases for write() (as is done in [WritableFileStream]([quota](https://storage.spec.whatwg.org/#storage-quota))): 

> If the operations modifying stream.[[[buffer]]](https://fs.spec.whatwg.org/#filesystemwritablefilestream-buffer) in the previous steps failed due to exceeding the [storage quota](https://storage.spec.whatwg.org/#storage-quota), [reject](https://webidl.spec.whatwg.org/#reject) p with a [QuotaExceededError](https://webidl.spec.whatwg.org/#quotaexceedederror) and abort, leaving stream.[[[buffer]]](https://fs.spec.whatwg.org/#filesystemwritablefilestream-buffer) unmodified.

> Note: [Storage quota](https://storage.spec.whatwg.org/#storage-quota) only applies to files stored in the [origin private file system](https://fs.spec.whatwg.org/#origin-private-file-system). However this operation could still fail for other files, for example if the disk being written to runs out of disk space.

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+   1. Set |tail| to a [=byte sequence=] containing the last
+      |oldSize| &minus; (|writePosition| + |data|'s [=byte sequence/length=]) bytes of |fileContents|.
+1. Let |newSize| be |head|'s [=byte sequence/length=] + |data|'s [=byte sequence/length=] + |tail|'s [=byte sequence/length=].
+1. If |newSize| &minus; |oldSize| exceeds the available [=storage quota=], throw a {{QuotaExceededError}}.
+1. Set [=this=].[=[[file]]=]'s [=file entry/binary data=] to the concatenation of |head|, |data| and |tail|.
+1. If the operations modifying the [=this=].[=[[file]]=]'s [=file entry/binary data=] in the previous steps
+   failed:
+    1. If there were partial writes and the number of bytes that were written from |data| is known,
+       return the number of bytes that were written from |data|.
+    1. Else return 0.
```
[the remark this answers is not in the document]
jesup: Well, in full disk cases write() will return less than the requested amount, and will have modified (and maybe extended) the file. (on linux)

a-sully:
> > See [this doc](http://doc/1lsQhTsfcVIeOW80dr467Auud_VCeAUv2ZOkC63oSyKo) for the justification behind the split sync/async interface.
> 
> I think you mean [this](https://docs.google.com/document/d/1lsQhTsfcVIeOW80dr467Auud_VCeAUv2ZOkC63oSyKo/edit?pli=1#)

Ah yes, sorry about that. I've updated the link in my comment.

index.bs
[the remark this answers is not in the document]
a-sully: > The implementation in Chrome currently returns `InvalidStateError` if you try to create an AccessHandle ... on a locked file

I would prefer to see `NoModificationAllowedError` used when the file is inaccessible (which is what this PR suggests), either due to locking by the API or because the file is not readable/writable on the underlying file system (which should only be relevant outside of the OPFS). I'll update the web platform tests accordingly.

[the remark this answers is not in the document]
annevk: I think one important thing to consider here is if callers are actually expected to branch on the different exception types. If not, I don't think there is much value in trying to distinguish the cases and it might even make certain optimizations harder.

index.bs
annevk: I think covering all of those scenarios would actually be really good and I'd consider them somewhat in scope as in that they can happen and user agents ideally behave in a consistent manner around them. Having said that, the current text seems hardly adequate for that? And depending on when these operations happen relative to each other, can you really obtain a meaningful number of modified bytes? If the file was removed you get its size back?

index.bs
annevk: That document explains why there is a synchronous API, but it doesn't explain the design decisions behind the particulars of the synchronous API. It also suggests the synchronous API is about to be deprecated (before it's even standardized!) which is interesting and not part of this PR. If we actually expect it to disappear at some point we want to be a lil clearer about that, but that also seems very strange as normally we provide APIs "forever".

index.bs
annevk: (A storage shelf is essentially the storage space for an "origin" (really storage key). https://storage.spec.whatwg.org/#model goes into this in more detail.)

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+### The {{FileSystemSyncAccessHandle/write()}} method ### {#api-filesystemsyncaccesshandle-write}
+
+<div class="note domintro">
+  : |handle| . {{FileSystemSyncAccessHandle/write()|write}}(|buffer|)
+  : |handle| . {{FileSystemSyncAccessHandle/write()|write}}(|buffer|, { {{FileSystemReadWriteOptions/at}} })
+  :: Writes the content of |buffer| into the file associated with |handle|, optionally at a given offset.
+</div>
+
+// TODO(fivedots): Figure out how to properly check the available storage quota (in this method and others) by passing the right storage shelf.
+// TODO(fivedots): Figure out mechanism to prevent concurrent IO operations e.g. doing write() while a truncate() is executing in parallel.
```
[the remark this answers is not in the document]
jesup: Right, but this action/error is not spec'd. Is that in fact the correct thing to do?  Or should it queue?   (i.e. dispatch a set of async operations, which queue, which is more typical async behavior on the web).

index.bs
```diff
@@ -314,6 +354,49 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method,
...
+<div algorithm>
+The <dfn method for=FileSystemFileHandle>createSyncAccessHandle()</dfn> method, when invoked, must run these steps:
+
+1. Let |result| be [=a new promise=].
+1. Run the following steps [=in parallel=]:
+  1. Let |access| be the result of running [=this=]'s [=FileSystemHandle/entry=]'s
+     [=entry/request access=] given "`readwrite`".
+     If that throws an exception, [=reject=] |result| with that exception and abort.
+  1. If |access| is not "{{PermissionState/granted}}",
+     reject |result| with a {{NotAllowedError}} and abort.
```
annevk, line 387: This is a comment that applies throughout: while you're "in parallel" you first need to queue a task and from that task you can manipulate state in the other thread, such as rejecting or resolving a promise.

You want to invoke HTML's "queue a global task" for this. (The Storage Standard hasn't been updated yet to account for this.) I suspect we want a task source that most storage APIs can share, e.g., "storage task source", though we could have one dedicated to this API I suppose, e.g., "file system task source". If we go with the former we'd need to add it to the Storage Standard. I can help with that.

index.bs
```diff
@@ -729,13 +812,18 @@ in a [=/Realm=] |realm|, perform the following steps:
 
          Note: It is expected that this atomically updates the contents of the file on disk
          being written to.
+
+      1. [=file entry/lock/release|Release the lock=] on |stream|.[=FileSystemWritableFileStream/[[file]]=].
       1. [=/Resolve=] |closeResult| with `undefined`.
```
jesup, line 817: Just wondering why close() returns undefined instead of success (or failure in the case of an error (close() *can* fail)

index.bs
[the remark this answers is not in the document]
a-sully: Whoops, I did mean that it throws, since there is no promise to reject!

We've since [updated the Chrome implementation](https://crrev.com/c/3657990) to support partial writes. The write method should only reject when nothing could be written

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+1. Run the following steps [=in parallel=]:
+  1. Let |oldSize| be the [=byte sequence/length=] of [=this=].[=[[file]]=]'s [=file entry/binary data=].
+  1. If |newSize| is larger than |oldSize|:
+     1. If |newSize| &minus; |oldSize| exceeds the available [=storage quota=], [=/reject=] |p|
+        with a {{QuotaExceededError}} and abort.
+     1. Set [=this=].[=[[file]]=]'s to a [=byte sequence=] formed by concatenating
+        |fileContents| with a [=byte sequence=]
+        containing |newSize| &minus; |oldSize| `0x00` bytes.
+     1. If the operations modifying the [=this=].[=[[file]]=]'s [=file entry/binary data=] in the previous steps
+        failed, [=/reject=] |p| with a {{InvalidStateError}} and abort.
```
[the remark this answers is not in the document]
a-sully: I agree that most of the actionable error codes would have thrown by now. That being said, do we have to specify what type of error is thrown here? Could this lack of actionability be an argument for throwing the most descriptive error we can? It seems unfortunate to throw all file errors into one bucket. This may not matter in the OPFS, but these file errors are more likely to crop up outside of OPFS, where having a more descriptive error message may help with debugging.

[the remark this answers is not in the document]
a-sully: After looking at this more closely, returning `InvalidStateError` actually DOES match the current implementation in this case. I agree with Emanuel that we should keep this simple and always throw `InvalidStateError`

index.bs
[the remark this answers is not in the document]
a-sully: SGTM! That matches our implementation :)

index.bs
fivedots: Oops, I missed this comment when replying to the last round, fixed!

index.bs
a-sully: Opening an issue about what to do when the file is modified by some other process (such as `rm -rf` and leaving a TODO in the spec sounds good to me

index.bs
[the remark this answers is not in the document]
a-sully: This will no longer be relevant once the API is made all-sync, so perhaps we can just delete this comment entirely?

index.bs
```diff
@@ -931,6 +1019,231 @@ steps:
...
+// TODO(fivedots): Figure out mechanism to prevent concurrent IO operations e.g. doing write() while a truncate() is executing in parallel.
+<div algorithm>
+The <dfn method for=FileSystemSyncAccessHandle>write(|buffer|, {{FileSystemReadWriteOptions}}: |options|)</dfn> method, when invoked, must run
+these steps:
+
+1. If [=this=].[=[[state]]=] is "`closed`", throw a {{InvalidStateError}}.
+1. Let |writePosition| be |options|.{{FileSystemReadWriteOptions/at}}.
+1. Let |fileContents| be a copy of [=this=].[=[[file]]=]'s [=file entry/binary data=].
+1. Let |oldSize| be |fileContents|'s [=byte sequence/length=].
+1. Let |data| be [=get a copy of the buffer source|a copy of=] |buffer|.
```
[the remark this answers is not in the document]
fivedots: After thinking it through a bit more, I'm not sure the direct reading mechanism used in that algorithm will work for us.

The approach I was considering was: set the file's data to |head| + a sequence of 0 the length of buffer + |tail|, and then directly filling in the 0's with the contents of buffer. The problem is that, in reality, implementations are likely going to directly modify the affected section of the file. In the case of a partial write it would be hard/slow to comply with the spec (e.g. if there is a partial write during the direct reading of the buffer, a section of the file should be left with the original 0's).

I've changed the language to be more purposely vague, similar to what we'd have to with flush and other operations that are close to the OS. WDYT?

index.bs
jesup: Yeah, I think this is ok

index.bs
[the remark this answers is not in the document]
jesup: ", at which point write order and the management partial write cannot be generally specified." -- I think this needs to be reworded to be clearer (and "the management partial write" doesn't make syntactical sense)
" return the number of bytes that were written from data." (for partial writes): 'data' isn't defined here; I think you mean 'buffer'.
We may want to add a note to developers to always check the return from write() to detect failures, since we throw in most cases, but for partial writes we return a smaller size.   Perhaps we should throw here as well, and perhaps provide some way to know how much was written in the failure case.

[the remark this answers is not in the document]
fivedots: I clarified this section, added a note for web developers above, and fixed the |data|<->|buffer| issue.

a-sully:
Friendly ping @jesup @annevk 

I'd like to get this PR approved and merged so we can move forward with making all the methods of the SyncAccessHandle sync :)

Thanks!

jesup:
@a-sully The WPT tests seem to assume you must pass 'at', and if you don't all reads start at  0.   This is ... odd, especially for implementing the equivalent to read(), which maintains an internal position in the file, and reads from there.   It means for almost all reads (or writes), you *must* pass at, which means you must track the expected position in the file - likely in a wrapper around syncaccesshandle.read()/write().   Emscriptem certainly can track the location in the file and insert at:nnnn always, but this seems like an odd choice for an API (and not great for performance, due to the extra parsing and validation and the basically forced extra kernel call to seek -- we could avoid the seek() call if we *also* track the position and elide if if it's a no-op, but that's not worth it either).

jesup:
Though maybe not legacy -- SyncAccessHandles came later.    Can (should?) we change this?   If you're writing a bunch of data, it's a whole lot more (unnecessary) bookkeeping and overhead.
```
   position += handle.write(data1, {at: position});
   position += handle.write(data2, {at: position});
```
But you can't quite do that, since if there's an error and a partial write, you won't notice it.  To be pedantic about catching errors, you need to do:
```
try {
  written = handle.write(data1, {at: position});
  if (written != data1.length()) { /* error */} else { postion += written;}
  written = handle.write(data2, {at: position});
  if (written != data2.length()) { /* error */} else { postion += written;}
} catch(e) { ... }
```

With classic posix-like read/write (with an assumed position update), it would be
```
  handle.write(data1);
  handle.write(data2);
```
or
```
try {
  if (handle.write(data1) != data1.length) { /* error */}
  if (handle.write(data2) != data2.length) { /* error */}
} catch(e) { ... }
```

index.bs
jesup: Ok

jesup:
We should probably merge this and then raise any points like the above as issues

jesup:
@tomayac Thoughts on at: ?

a-sully:
> We should probably merge this and then raise any points like the above as issues

I agree. Would you mind opening this in another issue? I have some thoughts here and it would be nice to have all the discussion about this in one place

jesup:
Editorial nit: in truncate(): "If the offset is smaller than offset, it remains unchanged."

jesup:
https://github.com/whatwg/fs/issues/49

jesup:
I can't approve since I don't have commit access, but go ahead.

Issue #14, opened 2022-03-08
Expand on "File object will likely be no longer readable"


a-sully:
_Migrated from https://github.com/WICG/file-system-access/issues/121 and https://github.com/WICG/file-system-access/issues/244_ (see https://github.com/whatwg/fs/issues/2)

### @foolip:

https://fs.spec.whatwg.org/#api-filesystemfilehandle-getfile says in domintro:

```
If the file on disk changes or is removed after this method is called, the returned File object will likely be no longer readable.
```

This isn't the normative text, but when the normative algorithm is written down will there really be any uncertainty to justify "likely" or can it be defined exactly what happens and when.

For a File that is no longer readable, does that mean that it's somehow neutered? Would lastModified reflect the change/removal time, or how would one tell that a File object is in this state?

### @mkruisselbrink:

The note there is mostly trying to describe what is already (not well) defined in FileAPI. I.e. this is the behavior for all File objects that represent actual files on disk. Unfortunately this is also not very well specified in FileAPI (see for example https://github.com/w3c/FileAPI/issues/47, but also https://github.com/w3c/FileAPI/issues/75). I definitely intend to fix that, but will need to fix the FileAPI side of things before I can fix this side.

### @mkruisselbrink:

We should define behavior of FileSystemFileHandle.getFile() when the underlying file no longer exists. Currently the spec is pretty much silent on this, we should probably make it clear/explicit that this should reject with a NotFoundError.

jesup:
So "snapshot state" ("Set f’s [snapshot state](https://w3c.github.io/FileAPI/#snapshot-state) to the current state of entry.") indicates that the data returned may not reflect later changes to the file; implying a read-into-memory (or copy) operation when the snapshot state is created.  Note that the definition of 'snapshot state' is somewhat vague and circular if you follow the links.
This is at odds with the comments in this issue and in the getFile definition.

annevk:
See the discussion in https://github.com/w3c/FileAPI/issues/47. @saschanaz has been doing some work in this area and it seems rather hard to completely prevent a `File` object from being readable after its backing file was modified on disk. For typical cases where the backing file's timestamp changes we can do it, but if that stays the change there's not really a way to do it.

jesup:
The wpt tests assume that using a blob after the original file has been removed will reject with NotFoundError; this isn't in the spec (and may not be always possible per above, and shouldn't be the case if my comment from June 15 is correct (that snapshot state implies the data is read or copied at creation time))

[...]

Issue #4, opened 2022-01-20
FileSystemDirectoryHandle.resolve is a very vague method name, consider renaming it


smaug----:
"resolve" doesn't hint too much what it is actually doing. One needs to read the algorithm to see what it actually returns.

a-sully:
I agree that the name is not great. Current usage is [still low](https://chromestatus.com/metrics/feature/timeline/popularity/3178) so I'm not necessarily opposed to a name change, but it would be a bit of a hassle to deprecate the current name. Do you think the benefits of renaming this method would be worth that hassle?

If so, the next question is what this should be named instead. Something like `getRelativePath()` or `getRelativePathComponents()` is more descriptive of what's happening here (since we don't actually return the path, but an array of path components). While we're at it, reversing the parameter order might make more sense...

Current: `dir.resolve(file)`

Alternatives:
- `file.getRelativePath(dir)`
- `file.getRelativePathComponents(dir)`
- `file.getPathRelativeTo(dir)`
- others?

poirierlouis:
Hello, I don't know if I'm trespassing by commenting here. Feel free to remove this if necessary.

> Do you think the benefits of renaming this method would be worth that hassle?

IMHO, it would be worth, as `resolve` is not particularly explicit. Someone discovering the API would need to read the documentation to be certain of what this method does. I believe alternatives proposed are more explicit when discovering this API, as 'path' and 'relative' makes direct sense in API's context.

I also like `file.getPathRelativeTo(dir)`.

> While we're at it, reversing the parameter order might make more sense...

Do you mean it would be called like this `dir.getPathRelativeTo(file)` / `dir.resolve(file)`?

annevk:
Unfortunately it seems the ball was dropped on this one. I don't see us renaming this at this point.

PR #30, opened 2022-05-24
Make explicit that FileSystemHandle entries map to paths

a-sully:
Be more explicit in describing how FileSystemHandles map to a path on disk, and the implications of this choice. Consider the following cases:

### Clearing site data
```
// Create file at /dir/file.txt.
const dir_entry = root.getDirectoryHandle('dir', { create: true });
const file_entry = dir_entry.getFileHandle('file.txt', { create: true });

// ======== Clear site data, clearing all data in the OPFS. ========

// file_entry still points to /dir/file.txt. The promise rejects with NotFoundError.
await dir_entry.getFile();

// Re-create the directory in the fresh OPFS.
const dir_entry = root.getDirectoryHandle('dir', { create: true });

// file_entry may be able to write /dir/file.txt, despite the file not being explicitly created
// Whether this works may depend on the user agent's implementation?
await file_entry.createWritable();
```

### Moving a file handle (see https://github.com/whatwg/fs/pull/10)
```
// Create file at /old/file.txt.
const dir_entry = root.getDirectoryHandle('old', { create: true });
const file_entry = dir_entry.getFileHandle('file.txt', { create: true });

// The file previously pointed to file at /new/file.txt now resides at /old/file.txt.
await dir_entry.move('new');

// file_entry still points to /old/file.txt. The promise rejects with NotFoundError.
await dir_entry.getFile();

// Restore the directory name. file_entry once again points to a valid path.
await dir_entry.move('old');

// file_entry still points to /old/file.txt. This call should succeed.
await file_entry.getFile();
```

### Removing a file handle (see https://github.com/whatwg/fs/pull/9)
```
// Create file at /old/file.txt.
const dir_entry = root.getDirectoryHandle('old', { create: true });
const file_entry = dir_entry.getFileHandle('file.txt', { create: true });

// Remove the file.
await file_entry.remove();

// We can still write to a removed file.
await file_entry.createWritable();
```

***
<a href="https://whatpr.org/fs/30.html" title="Last updated on May 26, 2022, 9:06 PM UTC (f55bd64)">Preview</a> | <a href="https://whatpr.org/fs/30/464b505...f55bd64.html" title="Last updated on May 26, 2022, 9:06 PM UTC (f55bd64)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -111,8 +111,9 @@ is left up to individual user-agent implementations.
 An [=/entry=] |a| is <dfn for="entry">the same as</dfn> an [=/entry=] |b| if |a| is equal to |b|, or
 if |a| and |b| are backed by the same file or directory on the local file system.
 
-Issue: TODO: Explain better how entries map to files on disk (multiple entries can map to the same file or
-directory on disk but an entry doesn't have to map to any file on disk).
+An [=/entry=] maps roughly to a path, but there is no guarantee this path exists on disk.
+For example, an [=/entry=] mapping to a path in the [=origin private file system=] will map to a path which
+no longer exists if site data is cleared.
 
 <div algorithm>
 To <dfn for="entry">resolve</dfn> an [=/entry=] |child| relative to a [=directory entry=] |root|,


index.bs
```diff
@@ -111,8 +111,9 @@ is left up to individual user-agent implementations.
 An [=/entry=] |a| is <dfn for="entry">the same as</dfn> an [=/entry=] |b| if |a| is equal to |b|, or
 if |a| and |b| are backed by the same file or directory on the local file system.
 
-Issue: TODO: Explain better how entries map to files on disk (multiple entries can map to the same file or
-directory on disk but an entry doesn't have to map to any file on disk).
+An [=/entry=] maps roughly to a path, but there is no guarantee this path exists on disk.
```
a-sully, line 114: @mkruisselbrink do you have a suggestion for a better way to word this?

I'm still hedging with the use of "roughly" here. I'm not sure how to toe the line in being as explicit as we can about what exactly the entry is, while allowing for entries to point to places other than disk (Drive, etc)

a-sully:
We've had a number of requests to expose the full path on the non-standardized portion of the API (https://github.com/WICG/file-system-access/issues/178, for example), but for privacy reasons I don't see us ever exposing a path for handles outside of the OPFS.

That being said, I don't think a full location is something we'd want or need to expose within the OPFS, either. A relative path is already available by calling `resolve()` with the handle and `navigator.storage.getDirectory()` so I'm not sure if I see the value of providing this on the object itself?

There's also the question of how this path would interact with Buckets. For example, multiple handles could have the same path but still be distinct files because they're in different buckets (unless the path includes the bucket somehow, which is a can of worms I assume we don't want to open). The advantage of using `resolve()` is that you have to explicitly pass in what the handle is relative to.

On that note, I'm open to renaming `resolve()` to something like `getRelativePath()` (see https://github.com/whatwg/fs/issues/4)

a-sully:
https://github.com/whatwg/fs/issues/14 is somewhat related

annevk:
I'm saying that we add it as a (private) member to the "entry" concept. Not that we'd expose it through a public API.

a-sully:
Ah I see. I did not realize we could spec a private member.

> There's also the question of how this path would interact with Buckets ... unless the path includes the bucket somehow, which is a can of worms I assume we don't want to open

It sounds like this member SHOULD include the bucket information. Assuming every new bucket has a unique ID (which seems like something we could spec?), this would handle the "clear site data" case above (or future "clear bucket" mechanism), since it would guarantee that handles become invalidated (well, at least always return `NotFoundError`) alongside their respective buckets (since the path/location will never be re-created).

Should we prefer "location" to "path", since this generalizes better to things like in-memory file systems and makes it clear that this is not a path relative to a bucket, but a location identifier which should include bucket information?

Overall this seems reasonable to me as long as we're okay with the implications for `move()` and `remove()` mentioned above. @mkruisselbrink any other thoughts here?

PR #65, opened 2022-10-26
Modernize algorithm language as per #64

### LINKED ISSUE #64: Modernize method and getter/setter definitions
As per Web IDL instead of

> The isSameEntry(other) method, when invoked, must run these steps:

we would write

> The isSameEntry(other) method steps are:

these days.

a-sully:
Editorial change. Fixes #64

Before: The <method> method, when invoked, must run these steps:

After: The <method> method steps are:

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -244,7 +244,7 @@ associated [=FileSystemHandle/entry=].
 </div>
 
 <div algorithm>
-The <dfn method for=FileSystemHandle>isSameEntry(|other|)</dfn> method, when invoked, must run these steps:
+The <dfn method for=FileSystemHandle>isSameEntry(|other|)</dfn> method steps are:
 
 1. Let |realm| be [=this=]'s [=relevant Realm=].
 1. Let |p| be [=a new promise=] in |realm|.
@@ -287,7 +287,7 @@ A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/entry=] must be a [=f
 </div>
 
 <div algorithm>
-The <dfn method for=FileSystemFileHandle>getFile()</dfn> method, when invoked, must run these steps:
+The <dfn method for=FileSystemFileHandle>getFile()</dfn> method steps are:
 
 1. Let |result| be [=a new promise=].
 1. Run the following steps [=in parallel=]:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -727,14 +724,12 @@ if (relative_path === null) {
 </div>
 
 <div algorithm>
-The <dfn method for=FileSystemDirectoryHandle>resolve(|possibleDescendant|)</dfn> method,
-when invoked, must return the result of [=entry/resolving=]
-|possibleDescendant|'s [=FileSystemHandle/entry=] relative to [=this=]'s [=FileSystemHandle/entry=].
+The <dfn method for=FileSystemDirectoryHandle>resolve(|possibleDescendant|)</dfn> method
+must return the result of [=entry/resolving=]|possibleDescendant|'s
+[=FileSystemHandle/entry=] relative to [=this=]'s [=FileSystemHandle/entry=].
```
a-sully, line 729: Just a heads up that this is a slightly change that doesn't exactly follow the pattern

annevk, line 729:
```suggestion
The <dfn method for=FileSystemDirectoryHandle>resolve(|possibleDescendant|)</dfn> method steps are
to return the result of [=entry/resolving=]|possibleDescendant|'s [=FileSystemHandle/entry=]
relative to [=this=]'s [=FileSystemHandle/entry=].
```

index.bs
```diff
@@ -727,14 +724,12 @@ if (relative_path === null) {
...
 
 </div>
 
-
-
```
annevk, line 737: Was this intentional?

index.bs
```diff
@@ -1219,8 +1206,7 @@ these steps:
 
 //TODO(fivedots): Figure out language to describe flushing the file at the OS level before closing the handle.
 <div algorithm>
-The <dfn method for=FileSystemSyncAccessHandle>close()</dfn> method, when invoked, must run
-these steps:
+The <dfn method for=FileSystemSyncAccessHandle>close()</dfn> method steps are:
```
annevk, line 1209: Since this is a single step you could inline this one as well.

annevk (commented):
Looks good.

Do you want to update the attributes (to use getter and setter steps) and constructors and such as well? Or rather in a new PR?

index.bs
a-sully: Yes, these are some extraneous newlines

index.bs
a-sully: Thanks for the suggestion!

a-sully:
> Do you want to update the attributes (to use getter and setter steps) and constructors and such as well? Or rather in a new PR?

I can make a new PR

PR #68, opened 2022-10-27
Editorial: modernize getters and setters

### LINKED ISSUE #64: Modernize method and getter/setter definitions

a-sully:
Fixes #64 

Only modernizes getters and setters. There are no constructors in this spec.

***
<a href="https://whatpr.org/fs/68.html" title="Last updated on Oct 28, 2022, 5:10 PM UTC (b722b4f)">Preview</a> | <a href="https://whatpr.org/fs/68/d64f799...b722b4f.html" title="Last updated on Oct 28, 2022, 5:10 PM UTC (b722b4f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -229,12 +229,11 @@ Their [=deserialization steps=], given |serialized| and |value| are:
   :: Returns the [=entry/name=] of the entry represented by |handle|.
 </div>
 
-The <dfn attribute for=FileSystemHandle>kind</dfn> attribute must
-return {{FileSystemHandleKind/"file"}} if the associated [=FileSystemHandle/entry=] is a [=file entry=],
+The <dfn attribute for=FileSystemHandle>kind</dfn> getter steps are to
+return {{FileSystemHandleKind/"file"}} if this is a [=file entry=],
 and return {{FileSystemHandleKind/"directory"}} otherwise.
 
-The <dfn attribute for=FileSystemHandle>name</dfn> attribute must return the [=entry/name=] of the
-associated [=FileSystemHandle/entry=].
+The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return the [=entry/name=] of this.
 
 ### The {{FileSystemHandle/isSameEntry()}} method ### {#api-filesystemhandle-issameentry}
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -299,9 +298,9 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method steps are:
   1. Let |f| be a new {{File}}.
   1. Set |f|'s <a spec=FileAPI>snapshot state</a> to the current state of |entry|.
   1. Set |f|'s underlying byte sequence to a copy of |entry|'s [=binary data=].
-  1. Initialize the value of |f|'s {{File/name}} attribute to |entry|'s [=entry/name=].
-  1. Initialize the value of |f|'s {{File/lastModified}} attribute to |entry|'s [=file entry/modification timestamp=].
-  1. Initialize the value of |f|'s {{Blob/type}} attribute to an [=implementation-defined=] value, based on for example |entry|'s [=entry/name=] or its file extension.
+  1. Set |f|.{{File/name}} to |entry|'s [=entry/name=].
+  1. Set |f|.{{File/lastModified}} to |entry|'s [=file entry/modification timestamp=].
+  1. Set |f|.{{Blob/type}} to an [=implementation-defined=] value, based on for example |entry|'s [=entry/name=] or its file extension.
```
a-sully, line 303: These setters follow the "Set f.name to entry's name" pattern as opposed to the "Set f's name to entry's name" pattern to match https://w3c.github.io/FileAPI/#dfn-file

That being said, lines 299 and 300 make a lot more sense in the current pattern (there's no ByteSequence definition to link to, AFAIK).... so maybe we should make this consistent and use the latter pattern everywhere here?

index.bs
```diff
@@ -229,12 +229,11 @@ Their [=deserialization steps=], given |serialized| and |value| are:
   :: Returns the [=entry/name=] of the entry represented by |handle|.
 </div>
 
-The <dfn attribute for=FileSystemHandle>kind</dfn> attribute must
-return {{FileSystemHandleKind/"file"}} if the associated [=FileSystemHandle/entry=] is a [=file entry=],
+The <dfn attribute for=FileSystemHandle>kind</dfn> getter steps are to
+return {{FileSystemHandleKind/"file"}} if this is a [=file entry=],
 and return {{FileSystemHandleKind/"directory"}} otherwise.
```
annevk, line 234:
```suggestion
The <dfn attribute for=FileSystemHandle>kind</dfn> getter steps are to return 
{{FileSystemHandleKind/"file"}} if [=this=] is a [=file entry=]; otherwise 
{{FileSystemHandleKind/"directory"}}.
```

index.bs
```diff
@@ -229,12 +229,11 @@ Their [=deserialization steps=], given |serialized| and |value| are:
...
 
-The <dfn attribute for=FileSystemHandle>name</dfn> attribute must return the [=entry/name=] of the
-associated [=FileSystemHandle/entry=].
+The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return the [=entry/name=] of this.
```
annevk, line 236:
```suggestion
The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=this=]'s [=entry/name=].
```

index.bs
annevk: I think what you have here currently is okay. The larger problem here is that the File API doesn't do a good job of defining the model of `File` object so you cannot manipulate the internal fields directly, which is what you'd want to do here.

annevk (approved):

PR #69, opened 2022-10-27
Editorial: Modernize algorithm steps language

a-sully:
Related to #64.

The following changes are made (mostly) via find-and-replace:

- "following steps" -> "these steps"
- "perform the" -> "run the"

***
<a href="https://whatpr.org/fs/69.html" title="Last updated on Oct 27, 2022, 11:24 PM UTC (9a267e7)">Preview</a> | <a href="https://whatpr.org/fs/69/d64f799...9a267e7.html" title="Last updated on Oct 27, 2022, 11:24 PM UTC (9a267e7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -93,8 +93,7 @@ a <dfn for="file entry">lock</dfn> (a string that may exclusively be "`open`", "
 and a <dfn for="file entry">shared lock count</dfn> (a number representing the number shared locks that are taken at a given point in time).
 
 <div algorithm>
-To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of "`exclusive`" or "`shared`" on a given [=file entry=] |file|,
-run the following steps:
+To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of "`exclusive`" or "`shared`" on a given [=file entry=] |file|:
 
 1. Let |lock| be the |file|'s [=file entry/lock=].
 1. Let |count| be the |file|'s [=file entry/shared lock count=].
@@ -116,7 +115,7 @@ run the following steps:
 
 <div algorithm>
 To <dfn for="file entry/lock">release</dfn> a [=file entry/lock=] on a given [=file entry=] |file|,
-run the following steps:
+run these steps:
 
 1. Let |lock| be the |file|'s associated [=file entry/lock=].
 1. Let |count| be the |file|'s [=file entry/shared lock count=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -788,19 +787,19 @@ Similarly, when piping a {{ReadableStream}} into a {{FileSystemWritableFileStrea
...
-1. Let |closeAlgorithm| be the following steps:
+1. Let |closeAlgorithm| be these steps:
    1. Let |closeResult| be [=a new promise=].
-   1. Run the following steps [=in parallel=]:
+   1. Run these steps [=in parallel=]:
       1. Let |access| be the result of running |file|'s [=entry/query access=] given "`readwrite`".
       1. If |access| is not "{{PermissionState/granted}}",
          reject |closeResult| with a {{NotAllowedError}} and abort.
-      1. Perform [=implementation-defined=] malware scans and safe browsing checks.
+      1. Run [=implementation-defined=] malware scans and safe browsing checks.
```
a-sully, line 802: Note: this change is not find-and-replace, but it seems to fit the pattern

index.bs
```diff
@@ -811,7 +810,7 @@ in a [=/Realm=] |realm|, perform the following steps:
       1. [=file entry/lock/release|Release the lock=] on |stream|.[=FileSystemWritableFileStream/[[file]]=].
       1. [=/Resolve=] |closeResult| with `undefined`.
    1. Return |closeResult|.
-1. Let |abortAlgorithm| be the following step:
+1. Let |abortAlgorithm| be these steps:
```
a-sully, line 813: Question: should one-step algorithms say "this step" or "these steps"?

index.bs
```diff
@@ -94,7 +94,7 @@ and a <dfn for="file entry">shared lock count</dfn> (a number representing the n
 
 <div algorithm>
 To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of "`exclusive`" or "`shared`" on a given [=file entry=] |file|,
-run the following steps:
+run these steps:
```
annevk, line 97: @domenic has been encouraging me to leave of "run these steps" entirely, when they occur at the end. I think it works well, especially for sentences starting with "To".

index.bs
```diff
@@ -811,7 +810,7 @@ in a [=/Realm=] |realm|, perform the following steps:
       1. [=file entry/lock/release|Release the lock=] on |stream|.[=FileSystemWritableFileStream/[[file]]=].
       1. [=/Resolve=] |closeResult| with `undefined`.
    1. Return |closeResult|.
-1. Let |abortAlgorithm| be the following step:
+1. Let |abortAlgorithm| be these steps:
       1. [=file entry/lock/release|Release the lock=] on |stream|.[=FileSystemWritableFileStream/[[file]]=].
```
annevk, line 814:
```suggestion
1. Let |abortAlgorithm| be this step: [=file entry/lock/release|release the lock=] on 
   |stream|.[=FileSystemWritableFileStream/[[file]]=].
```
We generally don't use dots to separate fields. It's almost always `'s `, but I see this is done here more often.

index.bs
annevk: Yeah. I'm not sure how good we are on "Run" vs "Perform" in general. Both seem okay and JavaScript uses "Perform" a lot, but the web platform has indeed more "Run".

index.bs
a-sully: Ah nice, that sounds good to me

PR #72, opened 2022-10-31
Editorial: Do not render TODO(fivedots) in the spec

### LINKED ISSUE #66: //TODO(fivedots) and variants
These should probably be changed to use HTML comments? They currently render as part of the standard.

a-sully:
Fixes #66.

Updates some TODOs to reference a corresponding issue while turning others into HTML comments so they do not render as part of the spec

***
<a href="https://whatpr.org/fs/72.html" title="Last updated on Nov 9, 2022, 5:04 PM UTC (7468203)">Preview</a> | <a href="https://whatpr.org/fs/72/abad896...7468203.html" title="Last updated on Nov 9, 2022, 5:04 PM UTC (7468203)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -331,15 +331,15 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method steps are:
      {{FileSystemSyncAccessHandle|FileSystemSyncAccessHandles}} for the entry, until the stream is closed.
 </div>
 
-See <a href=...>WICG/file-system-access issue 67...</a> for discussion around
-and desire for a "inPlace" mode for createWritable (where changes will be
-written to the actual underlying file as they are written to the writer, for
-example to support in-place modification of large files or things like
-databases). This is not currently implemented in Chrome. Implementing this is
-currently blocked on figuring out how to combine the desire to run malware
+See <a href=https://github.com/WICG/file-system-access/issues/67>WICG/file-system-access issue #67</a>
+for discussion around and desire for a "inPlace" mode for createWritable (where
+changes will be written to the actual underlying file as they are written to the
+writer, for example to support in-place modification of large files or things
+like databases). This is not currently implemented in Chrome. Implementing this
+is currently blocked on figuring out how to combine the desire to run malware
 checks with the desire to let websites make fast in-place modifications to
-existing large files. In-place writes are available for files in the
-[=origin private file system=] via the {{FileSystemSyncAccessHandle}} interface.
+existing large files. In-place writes are available for files in the [=origin
+private file system=] via the {{FileSystemSyncAccessHandle}} interface.
 
 <div algorithm>
 The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method steps are:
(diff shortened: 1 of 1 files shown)


a-sully:
marking draft since this should not include the other changes (sorry about that)

a-sully:
Question: there's a bunch of `Issue(##)` in the spec corresponding to [WICG/file-system-access](https://github.com/WICG/file-system-access) issues. Is there a standard convention for pointing to issues from another spec?

annevk:
Typically you'd use something along these lines:

> `<p class=XXX>See <a href=...>WICG/file-system-access issue #...</a> for ...`

annevk:
But we're probably at the point where we should move those kind of issues over if they are still relevant?

a-sully:
Thanks for the tip!

> But we're probably at the point where we should move those kind of issues over if they are still relevant?

Turns out they were already migrated (whoops). I've updated all `Issue(XX)` uses to point to this spec :)

index.bs
```diff
@@ -331,12 +331,15 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method steps are:
      {{FileSystemSyncAccessHandle|FileSystemSyncAccessHandles}} for the entry, until the stream is closed.
 </div>
 
-Issue(67): There has been some discussion around and desire for a "inPlace" mode for createWritable
-(where changes will be written to the actual underlying file as they are written to the writer, for
-example to support in-place modification of large files or things like databases). This is not
-currently implemented in Chrome. Implementing this is currently blocked on figuring out how to
-combine the desire to run malware checks with the desire to let websites make fast in-place
-modifications to existing large files.
+See <a href=...>WICG/file-system-access issue 67...</a> for discussion around
```
annevk, line 334: You do actually have to fill out the ... bits yourself. 😊

a-sully: This is why you don't try to multitask 🤦

index.bs
```diff
@@ -331,12 +331,15 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method steps are:
      {{FileSystemSyncAccessHandle|FileSystemSyncAccessHandles}} for the entry, until the stream is closed.
 </div>
 
-Issue(67): There has been some discussion around and desire for a "inPlace" mode for createWritable
-(where changes will be written to the actual underlying file as they are written to the writer, for
-example to support in-place modification of large files or things like databases). This is not
-currently implemented in Chrome. Implementing this is currently blocked on figuring out how to
-combine the desire to run malware checks with the desire to let websites make fast in-place
-modifications to existing large files.
+See <a href=https://github.com/WICG/file-system-access/issues/67>WICG/file-system-access issue #67</a>
```
annevk, line 334:
```suggestion
<p class=XXX>See <a href=https://github.com/WICG/file-system-access/issues/67>WICG/file-system-access issue #67</a>
```

annevk (commented):
The `Issue(...):` markup doesn't seem to be generating a link, but I think it's good enough. One nit left.

annevk (approved):

Issue #34, opened 2022-06-21
Request for multiple readers via access handles


alshamma:
Our web app downloads large application specific static data files. They range in the 10kb to several hundred kb size. We use a variety of approaches to cache this data. We would prefer to store them in OPFS, and, to be able to share them between multiple tabs of our app. For example, if the user opens two tabs to our app, we want to have a single copy of the data. We would like to download the data once, and then be able to read using the OPFS access handles from multiple tabs. We prefer access handles for their good performance.

jesup:
(I believe I know the answer to this):  Do you need the ability to seek within these files?   I presume performance reading (from WASM) is important here.

Adding an 'accessMode: read' to createSyncAccessHandle() would seem straightforward, and can leverage existing accessmode logic used currently for SyncAccessHandle and getWritableFileStream

alshamma:
Yes, we do need the ability to seek within these files. Your assumption about the importance of the performance of the read operation is correct. :-)

a-sully:
Is a read-only handle sufficient, or do you need a handle you can write to and read from (from multiple places) at the same time? To me, this becomes a more broad question around how we should handle file locking (#18).

#19 lays out a more general proposal that gives developers more control over locking behavior, among other things.

Taking this proposal a bit further makes me wonder whether we should allow things like multiple writers to the same file. An argument could be made that since this is intended to be a high-performance API, the browser should get out of the way and let the developer and underlying OS handle keeping file data consistent.

alshamma:
We don't need to do read/write simultaneously. We typically write once and read many.

tlively:
> Taking this proposal a bit further makes me wonder whether we should allow things like multiple writers to the same file. An argument could be made that since this is intended to be a high-performance API, the browser should get out of the way and let the developer and underlying OS handle keeping file data consistent.

From Emscripten's point of view, allowing arbitrarily many readers and writers would be ideal. We aim to emulate a normal POSIX environment to make bringing applications to the Web as easy as possible, and any differences imposed by Web APIs become problems our users (or we) have to work around. Our user's applications already handle the challenges of simultaneous file access on native platforms, so the Web platform trying to protect them from those challenges causes more problems than it solves.

szewai:
> > Taking this proposal a bit further makes me wonder whether we should allow things like multiple writers to the same file. An argument could be made that since this is intended to be a high-performance API, the browser should get out of the way and let the developer and underlying OS handle keeping file data consistent.
> 
[...]

PR #73, opened 2022-11-02
Require an exclusive lock in removeEntry()

a-sully:
For the sake of backwards compatibility, `removeEntry()` currently only requires acquiring a shared lock in Chromium (though this is not specified). This means that you cannot remove a file with an open Access Handle, but you _can_ remove a file with an open `WritableFileStream`.

The `removeEntry()` method should take an exclusive lock, matching #9. At least until we change the locking paradigm to not be based on "exclusive" and "shared" locks in #34.

   * All implementers have expressed support that this should match `remove()` in https://github.com/whatwg/fs/issues/34
   * …
   * Chromium: https://crbug.com/1254078
   * Gecko: …
   * WebKit: …

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -656,6 +656,11 @@ The <dfn method for=FileSystemDirectoryHandle>removeEntry(|name|, |options|)</df
   1. If |access| is not "{{PermissionState/granted}}",
      reject |result| with a {{NotAllowedError}} and abort.
 
+  1. Let |lockResult| be the result of [=file entry/lock/take|taking a lock=]
+     with "`exclusive`" on |entry|.
+  1. If |lockResult| is false, [=reject=] |result| with a
+     "{{NoModificationAllowedError}}" {{DOMException}} and abort.
+
   1. [=set/For each=] |child| of |entry|'s [=directory entry/children=]:
     1. If |child|'s [=entry/name=] equals |name|:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -656,6 +656,11 @@ The <dfn method for=FileSystemDirectoryHandle>removeEntry(|name|, |options|)</df
   1. If |access| is not "{{PermissionState/granted}}",
      reject |result| with a {{NotAllowedError}} and abort.
 
+  1. Let |lockResult| be the result of [=file entry/lock/take|taking a lock=]
+     with "`exclusive`" on |entry|.
```
a-sully, line 660: Note: we need to ensure that the new locking paradigm covers locked directories appropriately - we should not be able to _remove_ a directory if any contained files are locked, but presumably moving a directory with contained files that are locked is okay

a-sully: Filed #137

a-sully:
Update: `removeEntry()` has been taking an exclusive lock on Chromium browsers for a while now. See https://crbug.com/1254078

This PR is also incorrect as written, since the lock should not be on _the directory itself_ but on the child being removed. I'll update that before sending for review...

Anyways, I filed #138 to track updating updating the spec both for `removeEntry()` and `remove()`

PR #82, opened 2023-01-04
Make FileSystemReadWriteOptions.at optional

### LINKED ISSUE #81: Prose for dealing with dictionary members is wrong
> Let readStart be options["[at](https://fs.spec.whatwg.org/#dom-filesystemreadwriteoptions-at)"], if given; otherwise [this](https://webidl.spec.whatwg.org/#this)'s [file position cursor](https://fs.spec.whatwg.org/#filesystemsyncaccesshandle-file-position-cursor).

The problem here is that it's always given. That's how Web IDL works.

So it should probably be optional https://github.com/whatwg/fs/pull/76#issuecomment-1370485096 as suggests, but the reason for that is that otherwise it's pretty much always 0, unless explicitly given with a different value.

a-sully:
Fixes #81

***
<a href="https://whatpr.org/fs/82.html" title="Last updated on Jan 5, 2023, 9:04 PM UTC (8902347)">Preview</a> | <a href="https://whatpr.org/fs/82/730e2ad...8902347.html" title="Last updated on Jan 5, 2023, 9:04 PM UTC (8902347)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1079,7 +1079,9 @@ The <dfn method for=FileSystemSyncAccessHandle>read(|buffer|, {{FileSystemReadWr
 1. Let |bufferSize| be |buffer|'s [=byte length=].
 1. Let |fileContents| be [=this=].[=FileSystemSyncAccessHandle/[[file]]=]'s [=file entry/binary data=].
 1. Let |fileSize| be |fileContents|'s [=byte sequence/length=].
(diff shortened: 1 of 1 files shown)


annevk:
You want to use `[=map/exists=]` as https://dom.spec.whatwg.org/#event-flatten-more does. In the algorithm you can treat an IDL dictionary as a map, essentially.

annevk (approved):

PR #70, opened 2022-10-27
Editorial: Change exceptions to be more precise

a-sully:
Fixes #63

Before: {{InvalidStateError Error}}

After: "{{InvalidStateError Error}}" {{DOMException}}

***
<a href="https://whatpr.org/fs/70.html" title="Last updated on Nov 2, 2022, 7:49 AM UTC (4390bf5)">Preview</a> | <a href="https://whatpr.org/fs/70/abad896...4390bf5.html" title="Last updated on Nov 2, 2022, 7:49 AM UTC (4390bf5)">Diff</a>

PR #84, opened 2023-01-06
Throw TypeError if passed an offset not supported on the underlying OS

### LINKED ISSUE #31: Seek target range
the webidl has 
`[[EnforceRange](https://webidl.spec.whatwg.org/#EnforceRange)] required [unsigned long long](https://webidl.spec.whatwg.org/#idl-unsigned-long-long) at;`

However the WPT test of negative seek targets for SyncAccessHandles looks for dom NotSupportedError; the webidl spec says it should throw a TypeError.    Likely this is simply a bug in the WPT test

a-sully:
Fixes #31

***
<a href="https://whatpr.org/fs/84.html" title="Last updated on Jan 11, 2023, 9:41 PM UTC (5978da3)">Preview</a> | <a href="https://whatpr.org/fs/84/048a672...5978da3.html" title="Last updated on Jan 11, 2023, 9:41 PM UTC (5978da3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1081,6 +1081,8 @@ The <dfn method for=FileSystemSyncAccessHandle>read(|buffer|, {{FileSystemReadWr
 1. Let |readStart| be |options|["{{FileSystemReadWriteOptions/at}}"] if
    |options|["{{FileSystemReadWriteOptions/at}}"] [=map/exists=]; otherwise
    [=this=]'s [=FileSystemSyncAccessHandle/file position cursor=].
+1. If the underlying file system does not support reading from a file offset of
+   |readStart|, throw a {{TypeError}}.
 1. If |readStart| is larger than |fileSize|:
     1. Set [=this=]'s [=FileSystemSyncAccessHandle/file position cursor=] to |fileSize|.
     1. Return 0.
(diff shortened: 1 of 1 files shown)


a-sully:
Please let me know if you'd prefer this to be non-normative text, or if we should add that in addition to the algorithm changes

annevk:
I see that we also have "`throw an {{InvalidStateError}}`" which presumably needs to be:

> `[=throw=] an "{{InvalidStateError}}" {{DOMException}}`

Given that, should this be a JS `TypeError` or should this be of the `DOMException` variety as well?

And yeah, normally the non-normative description does call out that the algorithm can throw exceptions and which kinds.

a-sully:
> I see that we also have "`throw an {{InvalidStateError}}`" which presumably needs to be:
> 
> > `[=throw=] an "{{InvalidStateError}}" {{DOMException}}`

Huh, looks like I missed some in #70? Must've been lost while resolving merge conflicts or something...

> Given that, should this be a JS `TypeError` or should this be of the `DOMException` variety as well?

Prior discussion (https://github.com/whatwg/fs/issues/31#issuecomment-1148396305) suggested using `TypeError`, though I'm happy to make this a `DomException` if we've changed our mind now that we're seeing this in context with the rest of the algorithm

annevk (approved):
I don't have a strong opinion on what exception to use so I'll defer to you on that. Looks good modulo the nit.

Issue #83, opened 2023-01-06
Make FileSystemSyncAccessHandle.close algorithm more descriptive


youennf:
It is unclear whether https://fs.spec.whatwg.org/#api-filesystemsyncaccesshandle-close will wait for the lock to be released or will just set the handle state synchronously but will release the lock asynchronously.

annevk:
Yeah, the non-normative description does talk about release, but the normative description doesn't. So currently no lock would be released which seems bogus.

Since everything else is on this object is synchronous I would kinda expect it to wait for release to return?

a-sully:
> So currently no lock would be released which seems bogus.

Yikes

> Since everything else is on this object is synchronous I would kinda expect it to wait for release to return?

Correct. The lock should be released synchronously, before `close()` returns

I'll put up a PR to fix the algorithm soon (after #85 lands, for the sake of avoiding unnecessary merge conflicts)

szewai:
(Since this issue is about making `FileSystemSyncAccessHandle.close` more descriptive, I will just comment here:)

There is a wpt test [fs/FileSystemSyncAccessHandle-close.https.tentative.worker.html](https://wpt.fyi/results/fs/FileSystemSyncAccessHandle-close.https.tentative.worker.html?label=master&label=experimental&aligned) ensuring `close()` is idempotent and I think Chromium does this, but it is not in spec. It would be nice to make it clear.

a-sully:
> There is a wpt test [fs/FileSystemSyncAccessHandle-close.https.tentative.worker.html](https://wpt.fyi/results/fs/FileSystemSyncAccessHandle-close.https.tentative.worker.html?label=master&label=experimental&aligned) ensuring `close()` is idempotent and I think Chromium does this, but it is not in spec. It would be nice to make it clear.

Thanks for bringing this up, @szewai. And you're correct that `close()` is intended to be idempotent and is implemented as such in Chromium. Amusingly, the current spec language is actually correct on that front (`set this.state to "closed"` _is_ idempotent) but this will become _more obviously idempotent_ once we add language to release the lock :)

szewai:
> > There is a wpt test [fs/FileSystemSyncAccessHandle-close.https.tentative.worker.html](https://wpt.fyi/results/fs/FileSystemSyncAccessHandle-close.https.tentative.worker.html?label=master&label=experimental&aligned) ensuring `close()` is idempotent and I think Chromium does this, but it is not in spec. It would be nice to make it clear.
> 
[...]

PR #85, opened 2023-01-07
Describe flushing + don't flush on close

### LINKED ISSUE #71: Describe how flush() should interact with the underlying OS
This behavior is currently not well specified

a-sully:
Fixes #71

I'm happy to bikeshed the exact text for flushing in the comments.

Also updates the `close()` method's description to not mention flushing. I didn't realize this language existed, but it turns out _none_ (!) of the browsers actually flush the file during `close()`, which matches the behavior of most underlying OSes when closing a file descriptor anyways.

Fixing the `close()` algorithm (to address #83) will be done in a follow-up

***
<a href="https://whatpr.org/fs/85.html" title="Last updated on Jan 10, 2023, 9:58 PM UTC (cc64c12)">Preview</a> | <a href="https://whatpr.org/fs/85/048a672...cc64c12.html" title="Last updated on Jan 10, 2023, 9:58 PM UTC (cc64c12)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -370,7 +370,7 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method 
   : |handle| = await |fileHandle| . {{FileSystemFileHandle/createSyncAccessHandle()|createSyncAccessHandle}}()
   :: Returns a {{FileSystemSyncAccessHandle}} that can be used to read from/write to the file.
      Changes made through |handle| might be immediately reflected in the file represented by |fileHandle|.
-     To ensure the changes are reflected in this file, the handle can be flushed or closed.
+     To ensure the changes are reflected in this file, the handle can be flushed.
 
      Creating a {{FileSystemSyncAccessHandle}} [=file entry/lock/take|takes an exclusive lock=] on the
      [=FileSystemHandle/entry=] associated with |fileHandle|. This prevents the creation of
@@ -676,7 +676,6 @@ The <dfn method for=FileSystemDirectoryHandle>removeEntry(|name|, |options|)</df
       1. [=/Resolve=] |result| with `undefined`.
   1. [=/Reject=] |result| with a {{NotFoundError}}.
 1. Return |result|.
-
 
 </div>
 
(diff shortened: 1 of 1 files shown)


a-sully:
## Flush text bikshedding

I tried to make the text specific enough to be useful without prescribing implementation details or being too OS-specific (I assume we shouldn't explicitly mention `fdatasync()`, for example?)

"file content" is intentional here because it gives user agents the freedom to not flush file metadata if they so choose. I'm not sure if that should be made explicit?

index.bs
```diff
@@ -1211,23 +1209,31 @@ The <dfn method for=FileSystemSyncAccessHandle>getSize()</dfn> method steps are:
 <div algorithm>
 The <dfn method for=FileSystemSyncAccessHandle>flush()</dfn> method steps are:
 
-Issue(71): Fill in, after figuring out language to describe flushing at the OS level.
+1. If [=this=].[=[[state]]=] is "`closed`", throw an {{InvalidStateError}}.
+1. Attempt to transfer all cached modifications of the file's content the file
+   system's underlying storage device.
+
+   Note: This may be a no-op on some file systems, such as in-memory file
```
annevk, line 1216: Maybe start with "This is also known as flushing."

index.bs
```diff
@@ -1211,23 +1209,31 @@ The <dfn method for=FileSystemSyncAccessHandle>getSize()</dfn> method steps are:
...
 </div>
 
 <div algorithm>
 The <dfn method for=FileSystemSyncAccessHandle>close()</dfn> method steps are
 to set [=this=].[=[[state]]=] to "`closed`".
 
-Issue(71): Figure out language to describe flushing the file at the OS level before closing the handle.
+Note: This method does not guarantee that all file modifications will be
+immediately reflected in the underlying storage device. Call the
+{{FileSystemSyncAccessHandle/flush()}} method if you require this guarantee.
```
annevk, line 1236: first*

Right? Because `flush()` doesn't change the state nor releases.

a-sully:
Addressed all comments in the latest patch. Thank you for the review!

annevk (approved):
Thanks!

PR #95, opened 2023-01-18
Define a parallel queue for all file system operations

### LINKED ISSUE #74: Consider specifying a parallel queue for operations which acquire locks
Otherwise lock acquisition will race with other lock-acquiring methods. See https://github.com/whatwg/fs/pull/9#discussion_r1018153282

a-sully:
Fixes #74 by specifying a parallel queue for all file system operations, not just locking (see https://github.com/whatwg/fs/pull/9/files#r1109679577)

   * Chromium
   * WebKit
   * N/A (unless there's a way to deterministically test race conditions that I don't know about)
   * Chromium: ...
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/fs/95.html" title="Last updated on Jun 13, 2023, 4:32 PM UTC (ba5a295)">Preview</a> | <a href="https://whatpr.org/fs/95/8360b75...ba5a295.html" title="Last updated on Jun 13, 2023, 4:32 PM UTC (ba5a295)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -67,7 +67,7 @@ algorithm, which takes "`read`" or "`readwrite`" <var ignore>mode</var> and
 returns a [=/file system access result=].
 Unless specified otherwise it returns a [=/file system access result=] with a
 [=file system access result/permission state=] of "{{PermissionState/denied}}"
-and with a [=file system access result/error name=] of « the empty string ».
+and with a [=file system access result/error name=] of the empty string.
 
 Each [=/file system entry=] has an associated
 <dfn for="file system entry" id=entry-request-access>request access</dfn>
@@ -75,7 +75,7 @@ algorithm, which takes "`read`" or "`readwrite`" <var ignore>mode</var> and
 returns a [=/file system access result=].
 Unless specified otherwise it returns a [=/file system access result=] with a
 [=file system access result/permission state=] of "{{PermissionState/denied}}"
-and with a [=file system access result/error name=] of « the empty string ».
+and with a [=file system access result/error name=] of the empty string.
 
 A <dfn export>file system access result</dfn> is a [=struct=] encapsulating the
 result of [=file system entry/query access|querying=] or
@@ -85,10 +85,13 @@ It has the following [=struct/items=]:
 : <dfn for="file system access result">permission state</dfn>
 :: A {{PermissionState}}
 : <dfn for="file system access result">error name</dfn>
-:: A [=string=] which must be either « the empty string » or an
+:: A [=string=] which must be either the empty string or an
    [=exception/error name=] listed in the [=error names table=].
-   If [=file system access result/permission state=] is
-   "{{PermissionState/granted}}" this must be « the empty string »
+   Iff [=file system access result/permission state=] is
+   "{{PermissionState/granted}}" this will be the empty string.
+   It is expected that in most cases when
+   [=file system access result/permission state=] is not
+   "{{PermissionState/granted}}", this will be "{{NotAllowedError}}".
 
 <p class=warning> Dependent specifications may consider this API a
 [=powerful feature=]. However, unlike other [=powerful features=] whose
@@ -98,7 +101,7 @@ algorithms must run [=in parallel=] on the [=file system queue=] and are
 therefore not allowed to throw. Instead, the caller is expected to
 [=queue a storage task=] to [=/reject=], as appropriate,
 should these algorithms return an [=file system access result/error name=]
-other than « the empty string ».
+other than the empty string.
 
 Note: Implementations that only implement this specification and not dependent
 specifications do not need to bother implementing [=/file system entry=]'s
@@ -477,12 +480,9 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method steps are:
 
   1. [=Queue a storage task=] with |global| to run these steps:
     1. If |accessResult|'s [=file system access result/permission state=]
-       is not "{{PermissionState/granted}}":
-      1. Set |accessErrorName| to |accessResult|'s
-         [=file system access result/error name=] if it is not
-         « the empty string »; otherwise "{{NotAllowedError}}".
-      1. [=/Reject=] |result| with an |accessErrorName| {{DOMException}} and
-         abort these steps.
+       is not "{{PermissionState/granted}}", [=/reject=] |result| with a
+       {{DOMException}} of |accessResult|'s
+       [=file system access result/error name=] and abort these steps.
 
     1. If |entry| is null, [=/reject=] |result| with a
        "{{NotFoundError}}" {{DOMException}} and abort these steps.
@@ -550,12 +550,10 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method 
   1. Let |accessResult| be the result of running |entry|'s
      [=file system entry/request access=] given "`readwrite`".
   1. If |accessResult|'s [=file system access result/permission state=]
-     is not "{{PermissionState/granted}}":
-    1. Set |accessErrorName| to |accessResult|'s
-       [=file system access result/error name=] if it is not
-       « the empty string »; otherwise "{{NotAllowedError}}".
-    1. [=Queue a storage task=] with |global| to [=/reject=] |result| with an
-       |accessErrorName| {{DOMException}} and abort these steps.
+     is not "{{PermissionState/granted}}", [=queue a storage task=] with
+     |global| to [=/reject=] |result| with a {{DOMException}} of
+     |accessResult|'s [=file system access result/error name=] and
+     abort these steps.
 
   1. If |entry| is `null`, [=queue a storage task=] with |global| to [=/reject=]
      |result| with a "{{NotFoundError}}" {{DOMException}} and abort these steps.
@@ -612,12 +610,10 @@ The <dfn method for=FileSystemFileHandle>createSyncAccessHandle()</dfn> method s
   1. Let |accessResult| be the result of running |entry|'s
      [=file system entry/request access=] given "`readwrite`".
   1. If |accessResult|'s [=file system access result/permission state=]
-     is not "{{PermissionState/granted}}":
-    1. Set |accessErrorName| to |accessResult|'s
-       [=file system access result/error name=] if it is not
-       « the empty string »; otherwise "{{NotAllowedError}}".
-    1. [=Queue a storage task=] with |global| to [=/reject=] |result| with an
-       |accessErrorName| {{DOMException}} and abort these steps.
+     is not "{{PermissionState/granted}}", [=queue a storage task=] with
+     |global| to [=/reject=] |result| with a {{DOMException}} of
+     |accessResult|'s [=file system access result/error name=] and
+     abort these steps.
 
   1. If |entry| is `null`, [=queue a storage task=] with |global| to [=/reject=]
      |result| with a "{{NotFoundError}}" {{DOMException}} and abort these steps.
@@ -727,20 +723,9 @@ Issue(15): In the future we might want to add arguments to the async iterable de
 support for example recursive iteration.
 
 <div algorithm="iterator initialization">
-The [=asynchronous iterator initialization steps=] for a {{FileSystemDirectoryHandle}} |handle|
+The [=asynchronous iterator initialization steps=] for a
+{{FileSystemDirectoryHandle}} <var ignore>handle</var>
 and its async iterator |iterator| are:
-
-1. [=Enqueue the following steps=] to the [=file system queue=]:
-  1. Let |entry| be the result of [=locating an entry=]
-     given |handle|'s [=FileSystemHandle/locator=].
-  1. Let |accessResult| be the result of running |entry|'s
-     [=file system entry/query access=] given "`read`".
-  1. If |accessResult|'s [=file system access result/permission state=]
-     is not "{{PermissionState/granted}}":
-    1. Set |accessErrorName| to |accessResult|'s
-       [=file system access result/error name=] if it is not
-       « the empty string »; otherwise "{{NotAllowedError}}".
-    1. [=Throw=] an |accessErrorName| {{DOMException}}.
 
 1. Set |iterator|'s <dfn for="FileSystemDirectoryHandle-iterator">past results</dfn> to an empty [=/set=].
 
@@ -760,12 +745,9 @@ and its async iterator |iterator|:
   1. [=Queue a storage task=] with |handle|'s [=relevant global object=] to
      run these steps:
     1. If |accessResult|'s [=file system access result/permission state=]
-       is not "{{PermissionState/granted}}":
-      1. Set |accessErrorName| to |accessResult|'s
-         [=file system access result/error name=] if it is not
-         « the empty string »; otherwise "{{NotAllowedError}}".
-      1. [=/Reject=] |promise| with an |accessErrorName| {{DOMException}}
-         and return |promise|.
+       is not "{{PermissionState/granted}}", [=/reject=] |promise| with a
+       {{DOMException}} of |accessResult|'s
+       [=file system access result/error name=] and abort these steps.:
 
     1. If |directory| is `null`, [=/reject=] |result| with a
        "{{NotFoundError}}" {{DOMException}} and abort these steps.
@@ -849,12 +831,9 @@ The <dfn method for=FileSystemDirectoryHandle>getFileHandle(|name|, |options|)</
 
   1. [=Queue a storage task=] with |global| to run these steps:
     1. If |accessResult|'s [=file system access result/permission state=]
-       is not "{{PermissionState/granted}}":
-      1. Set |accessErrorName| to |accessResult|'s
-         [=file system access result/error name=] if it is not
-         « the empty string »; otherwise "{{NotAllowedError}}".
-      1. [=/Reject=] |result| with an |accessErrorName| {{DOMException}} and
-         abort these steps.
+       is not "{{PermissionState/granted}}", [=/reject=] |result| with a
+       {{DOMException}} of |accessResult|'s
+       [=file system access result/error name=] and abort these steps.
 
     1. If |entry| is `null`, [=/reject=] |result| with a
        "{{NotFoundError}}" {{DOMException}} and abort these steps.
@@ -934,12 +913,9 @@ The <dfn method for=FileSystemDirectoryHandle>getDirectoryHandle(|name|, |option
 
   1. [=Queue a storage task=] with |global| to run these steps:
     1. If |accessResult|'s [=file system access result/permission state=]
-       is not "{{PermissionState/granted}}":
-      1. Set |accessErrorName| to |accessResult|'s
-         [=file system access result/error name=] if it is not
-         « the empty string »; otherwise "{{NotAllowedError}}".
-      1. [=/Reject=] |result| with an |accessErrorName| {{DOMException}} and
-         abort these steps.
+       is not "{{PermissionState/granted}}", [=/reject=] |result| with a
+       {{DOMException}} of |accessResult|'s
+       [=file system access result/error name=] and abort these steps.
 
     1. If |entry| is `null`, [=/reject=] |result| with a
        "{{NotFoundError}}" {{DOMException}} and abort these steps.
@@ -1010,12 +986,9 @@ The <dfn method for=FileSystemDirectoryHandle>removeEntry(|name|, |options|)</df
 
   1. [=Queue a storage task=] with |global| to run these steps:
     1. If |accessResult|'s [=file system access result/permission state=]
-       is not "{{PermissionState/granted}}":
-      1. Set |accessErrorName| to |accessResult|'s
-         [=file system access result/error name=] if it is not
-         « the empty string »; otherwise "{{NotAllowedError}}".
-      1. [=/Reject=] |result| with an |accessErrorName| {{DOMException}} and
-         abort these steps.
+       is not "{{PermissionState/granted}}", [=/reject=] |result| with a
+       {{DOMException}} of |accessResult|'s
+       [=file system access result/error name=] and abort these steps.
 
     1. If |entry| is `null`, [=/reject=] |result| with a
        "{{NotFoundError}}" {{DOMException}} and abort these steps.
@@ -1172,13 +1145,9 @@ given a [=file entry=] |file| in a [=/Realm=] |realm|:
       1. [=Queue a storage task=] with |file|'s [=relevant global object=]
          to run these steps:
         1. If |accessResult|'s [=file system access result/permission state=]
-           is not "{{PermissionState/granted}}":
-          1. Set |accessErrorName| to |accessResult|'s
-             [=file system access result/error name=] if it is not
-             « the empty string »; otherwise "{{NotAllowedError}}".
-          1. [=/Reject=] |closeResult| with an
-             |accessErrorName| {{DOMException}} and abort these steps.
-
+           is not "{{PermissionState/granted}}", [=/reject=] |closeResult|
+           with a {{DOMException}} of |accessResult|'s
+           [=file system access result/error name=] and abort these steps.
         1. Run [=implementation-defined=] malware scans and safe browsing checks.
            If these checks fail, [=/reject=] |closeResult| with an
            "{{AbortError}}" {{DOMException}} and abort these steps.
@@ -1229,12 +1198,9 @@ runs these steps:
   1. [=Queue a storage task=] with |stream|'s [=relevant global object=] to
      run these steps:
     1. If |accessResult|'s [=file system access result/permission state=]
-       is not "{{PermissionState/granted}}":
-      1. Set |accessErrorName| to |accessResult|'s
-         [=file system access result/error name=] if it is not
-         « the empty string »; otherwise "{{NotAllowedError}}".
-      1. [=/Reject=] |p| with an |accessErrorName| {{DOMException}} and
-         abort these steps.
+       is not "{{PermissionState/granted}}", [=/reject=] |p| with a
+       {{DOMException}} of |accessResult|'s
+       [=file system access result/error name=] and abort these steps.
 
     1. Let |command| be |input|.{{WriteParams/type}} if |input| is a {{WriteParams}},
        and {{WriteCommandType/"write"}} otherwise.
@@ -1705,7 +1671,7 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
      a [=/file system access result=]
      with a [=file system access result/permission state=]
      of "{{PermissionState/granted}}" and
-     with a [=file system access result/error name=] of « the empty string ».
+     with an [=file system access result/error name=] of the empty string.
   1. Set |dir|'s [=file system entry/name=] to the empty string.
   1. Set |dir|'s [=directory entry/children=] to an empty [=/set=].
   1. Set |map|["root"] to |dir|.
--- a/index.bs
+++ b/index.bs
@@ -94,24 +94,29 @@ A <dfn export id=file>file entry</dfn> additionally consists of
 a <dfn for="file entry">lock</dfn> (a string that may exclusively be "`open`", "`taken-exclusive`" or "`taken-shared`")
 and a <dfn for="file entry">shared lock count</dfn> (a number representing the number shared locks that are taken at a given point in time).
 
+The <dfn id="file-system-lock-queue">file system lock queue</dfn> is a
+[=parallel queue=] to be used for all <a for=/>tasks</a> involving a
+[=file entry/lock=].
+
 <div algorithm>
 To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of "`exclusive`" or "`shared`" on a given [=file entry=] |file|:
 
 1. Let |lock| be the |file|'s [=file entry/lock=].
-1. Let |count| be the |file|'s [=file entry/shared lock count=].
-1. If |value| is "`exclusive`":
-   1. If |lock| is "`open`":
-     1. Set lock to "`taken-exclusive`".
-     1. Return true.
-1. If |value| is "`shared`":
-   1. If |lock| is "`open`":
-     1. Set |lock| to "`taken-shared`".
-     1. Set |count| to 1.
-     1. Return true.
-   1. Otherwise, if |lock| is "`taken-shared`":
-     1. Increase |count| by one.
-     1. Return true.
-1. Return false.
+1. [=Enqueue the following steps=] to the [=file system lock queue=]:
+  1. Let |count| be the |file|'s [=file entry/shared lock count=].
+  1. If |value| is "`exclusive`":
+     1. If |lock| is "`open`":
+       1. Set lock to "`taken-exclusive`".
+       1. Return true.
+  1. If |value| is "`shared`":
+     1. If |lock| is "`open`":
+       1. Set |lock| to "`taken-shared`".
+       1. Set |count| to 1.
+       1. Return true.
+     1. Otherwise, if |lock| is "`taken-shared`":
+       1. Increase |count| by one.
+       1. Return true.
+  1. Return false.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


a-sully:
cc @jesup @szewai

index.bs
```diff
@@ -90,24 +90,28 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
+       1. Return true.
+  1. If |value| is "`shared`":
+     1. If |lock| is "`open`":
+       1. Set |lock| to "`taken-shared`".
+       1. Set |count| to 1.
+       1. Return true.
+     1. Otherwise, if |lock| is "`taken-shared`":
+       1. Increase |count| by one.
+       1. Return true.
+  1. Return false.
```
annevk, line 114: This doesn't really work. These steps end up running in another conceptual thread. So the return values are lost.

You need to communicate back to the main thread through tasks.

You also need to actually start the parallel queue as otherwise it won't be executing its steps. See https://html.spec.whatwg.org/multipage/infrastructure.html#starting-a-new-parallel-queue.

(Note that the example there mistakenly doesn't queue tasks. That's a bug that still needs to be fixed.)

a-sully: Ah, I see now. I made the mistake of basing this off of the example :)

I've sketched out what lock acquisition + release could look like on a couple methods. Please let me know if there's a better way you'd recommend doing this. Once it looks good, I'll copy this pattern to all the relevant methods

Some thoughts:
- Instead of passing the promise directly, should the "take a lock" algorithm be passed an "onFailure" algorithm (which could then reject the promise)? Right now we always reject if the lock acquisition fails, but presumably we could add a method that doesn't
- Is there a better way to do this? I found https://github.com/whatwg/infra/issues/181, but that doesn't seem to have gone anywhere
- Is it correct to have `onLockReleasedAlgorithm` be an algorithm, or should it be a task?

index.bs
```diff
@@ -90,37 +90,51 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
+     1. If |lock| is "`open`":
+       1. Set |lock| to "`taken-shared`".
+       1. Set |count| to 1.
+       1. [=Queue a task=] to |replyTaskSource| to run |onLockTakenAlgorithm|
+          and return.
+     1. Otherwise, if |lock| is "`taken-shared`":
+       1. Increase |count| by one.
+       1. [=Queue a task=] to |replyTaskSource| to run |onLockTakenAlgorithm|
+          and return.
+  1. [=Reject=] |p| with a "{{NoModificationAllowedError}}" {{DOMException}}.
```
annevk, line 120: You have to queue a task for this as well. But maybe it makes more sense that you don't pass the promise in to this algorithm and you have a single callback algorithm that you then pass success/failure to or some such?

index.bs
```diff
@@ -90,37 +90,51 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
+"`exclusive`" or "`shared`" on a given [=file entry=] |file|, with promise |p|,
+algorithm |onLockTakenAlgorithm|, and [=task source=] |replyTaskSource|:
+
+1. [=Enqueue the following steps=] to the [=file system lock queue=]:
+  1. Let |lock| be the |file|'s [=file entry/lock=].
+  1. Let |count| be the |file|'s [=file entry/shared lock count=].
+  1. If |value| is "`exclusive`":
+     1. If |lock| is "`open`":
+       1. Set lock to "`taken-exclusive`".
+       1. [=Queue a task=] to |replyTaskSource| to run |onLockTakenAlgorithm|
```
annevk, line 108: We want to be using "queue a global task". Passing a global across the "in parallel" boundary is generally accepted, assuming you don't start pulling state of it.

I also don't think we need _replyTaskSoruce_. We can just always use the storage task source.

annevk (commented):
This looks better, but not quite there yet. I'll be out next week, but happy to look at this again in February.

index.bs
annevk: 1. Yeah. Or a single algorithm that invoke with an argument to let it know success/failure.
2. Not currently.
3. No that's correct. The task is the eventual wrapper for the algorithm and some other state.

index.bs
a-sully: Ack

index.bs
a-sully: Makes sense! The original algorithm returned true or false, but it now runs the passed-in algorithm with "`success`" and "`failure`" explicitly, which seems more clear. Let me know if there's a better way to do this!

index.bs
a-sully: Changed to pass only a single algorithm. Thanks for the suggestion!

index.bs
```diff
@@ -1249,8 +1272,10 @@ The <dfn method for=FileSystemSyncAccessHandle>close()</dfn> method steps are:
 
 1. If [=this=]'s [=[[state]]=] is "`closed`", return.
 1. Set [=this=]'s [=[[state]]=] to "`closed`".
+1. Let |onLockReleasedAlgorithm| be an algorithm that does nothing.
 1. [=file entry/lock/release|Release the lock=] on
-   [=this=]'s [=FileSystemSyncAccessHandle/[[file]]=].
+   [=this=]'s [=FileSystemSyncAccessHandle/[[file]]=] with |onLockReleasedAlgorithm|.
+1. Wait for |onLockReleasedAlgorithm| to run.
```
a-sully, line 1278: Now that releasing a lock is async, presumably we can't just post a task to release the lock and return. Here's my attempt at saying "wait until the lock is released to return from `close()` - let me know if there's a better way to do this. Thanks!

annevk: I think it would be slightly clearer for _onLockReleasedAlgorithm_ to flip a boolean.

And then instead of wait you want pause as per https://html.spec.whatwg.org/multipage/webappapis.html#pause because we're inside an actual task and not in parallel.

(And also please ensure to never introduce such a thing outside of workers!)

index.bs
```diff
@@ -90,37 +90,51 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
+"`exclusive`" or "`shared`" on a given [=file entry=] |file| and with
+algorithm |resultAlgorithm|:
+
+1. [=Enqueue the following steps=] to the [=file system lock queue=]:
+  1. Let |lock| be the |file|'s [=file entry/lock=].
+  1. Let |count| be the |file|'s [=file entry/shared lock count=].
+  1. If |value| is "`exclusive`":
+     1. If |lock| is "`open`":
+       1. Set lock to "`taken-exclusive`".
+       1. [=Queue a global task=] on the [=storage task source=] to run
```
annevk, line 108: We want queue a storage task here and elsewhere. It seems like _file_'s relevant global object would be the correct global to use.

index.bs
```diff
@@ -90,37 +90,51 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
+          |resultAlgorithm| with a value of "`success`" and return.
+  1. If |value| is "`shared`":
+     1. If |lock| is "`open`":
+       1. Set |lock| to "`taken-shared`".
+       1. Set |count| to 1.
+       1. [=Queue a global task=] on the [=storage task source=] to run
+          |resultAlgorithm| with a value of "`success`" and return.
+     1. Otherwise, if |lock| is "`taken-shared`":
+       1. Increase |count| by one.
```
annevk, line 117:
```suggestion
       1. Increase |count| by 1.
```

index.bs
```diff
@@ -90,37 +90,51 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
-  1. If |count| is 0, set |lock| to "`open`".
-1. Otherwise, set |lock| to "`open`".
+To <dfn for="file entry/lock">release</dfn> a [=file entry/lock=] on a given
+[=file entry=] |file| and with an optional algorithm |onLockReleasedAlgorithm|:
+
+1. [=Enqueue the following steps=] to the [=file system lock queue=]:
+  1. Let |lock| be the |file|'s associated [=file entry/lock=].
+  1. Let |count| be the |file|'s [=file entry/shared lock count=].
+  1. If |lock| is "`taken-shared`":
+    1. Decrease |count| by one.
```
annevk, line 133:
```suggestion
    1. Decrease |count| by 1.
```

index.bs
```diff
@@ -353,13 +367,17 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method
   1. If |access| is not "{{PermissionState/granted}}",
      [=reject=] |result| with a "{{NotAllowedError}}" {{DOMException}} and abort.
```
annevk, line 368: This needs a task.

annevk (commented):
Mainly lots of nits, but this looks to be heading in the right direction to me!

index.bs
a-sully: Just to clarify, you're suggesting that the promise rejection should happen on some task source, not the `[=file system entry/request access=]` algorithm - correct?

This pattern (rejecting a promise from a step run in parallel) is repeated many times throughout this file. Would it be alright to fix these in a follow-up?

Also, just to confirm, is the promise rejection in the `lockResultAlgorithm` steps below fine as-is because it's already running on the storage task source? Or does that also need to queue a task?

Side note: while looking at this, I realized that `[=a new promise=]` is supposed to take a Realm. This appears to only be done correctly [in one method](https://fs.spec.whatwg.org/#dom-filesystemhandle-issameentry) for this spec. I can also fix this in a follow-up...

index.bs
a-sully: Aha! I wasn't aware we could pull file's relevant global object like that. I think this is right now. Sorry for the confusion!

index.bs
a-sully: > I think it would be slightly clearer for _onLockReleasedAlgorithm_ to flip a boolean.

Done!

> And then instead of wait you want pause as per https://html.spec.whatwg.org/multipage/webappapis.html#pause because we're inside an actual task and not in parallel.
> 
> (And also please ensure to never introduce such a thing outside of workers!)

You betcha :)

index.bs
a-sully: > This pattern (rejecting a promise from a step run in parallel) is repeated many times throughout this file. Would it be alright to fix these in a follow-up?

For an idea of what this could look like: https://github.com/whatwg/fs/pull/9/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R273-R276

annevk: Yeah, I guess a follow-up is okay.

And yeah, if you already have a task you're not in parallel and you can resolve promises as you wish. (Though sometimes you might still want to queue a task, it really depends on what kind of ordering you are after.)

index.bs
```diff
@@ -90,37 +90,54 @@ A <dfn export id=file>file entry</dfn> additionally consists of
 a <dfn for="file entry">lock</dfn> (a string that may exclusively be "`open`", "`taken-exclusive`" or "`taken-shared`")
 and a <dfn for="file entry">shared lock count</dfn> (a number representing the number shared locks that are taken at a given point in time).
 
+A user agent has an associated <dfn>file system lock queue</dfn> which is the
+result of [=starting a new parallel queue=]. This queue is to be used for all
+[=tasks=] involving a [=file entry/lock=].
```
annevk, line 95: s/tasks/operations/? (The tasks are not the essential part after all, they're just for communicating results.)

index.bs
```diff
@@ -90,37 +90,54 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
-1. If |value| is "`shared`":
-   1. If |lock| is "`open`":
-     1. Set |lock| to "`taken-shared`".
-     1. Set |count| to 1.
-     1. Return true.
-   1. Otherwise, if |lock| is "`taken-shared`":
-     1. Increase |count| by one.
-     1. Return true.
-1. Return false.
+To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of
```
annevk, line 98: Follow-up: maybe it would be clearer if we called this "take a lock". The current phrasing kinda gives the impression lock might be an argument.

index.bs
```diff
@@ -90,37 +90,54 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
+
+1. [=Enqueue the following steps=] to the [=file system lock queue=]:
+  1. Let |lock| be the |file|'s [=file entry/lock=].
+  1. Let |count| be the |file|'s [=file entry/shared lock count=].
+  1. If |value| is "`exclusive`":
+     1. If |lock| is "`open`":
+       1. Set lock to "`taken-exclusive`".
+       1. [=Queue a storage task=] given |file|'s [=relevant global object=] and
+          |resultAlgorithm| with a value of "`success`".
+       1. Return.
```
annevk, line 110: We use "Abort these steps" in "in parallel" algorithms to not confuse readers they might actually return something useful to someone.

index.bs
```diff
@@ -353,13 +370,17 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method
...
-  1. Let |stream| be the result of [=create a new FileSystemWritableFileStream|creating a new FileSystemWritableFileStream=]
-     for |entry| in [=this=]'s [=relevant realm=].
-  1. If |options|.{{FileSystemCreateWritableOptions/keepExistingData}} is true:
-    1. Set |stream|.[=[[buffer]]=] to a copy of |entry|'s [=file entry/binary data=].
-  1. [=/Resolve=] |result| with |stream|.
+  1. Let |lockResultAlgorithm| be an algorithm which takes a |lockResult| and
+     runs these steps:
+    1. If |lockResult| is "`failure`", [=reject=] |result| with a
+       "{{NoModificationAllowedError}}" {{DOMException}} and abort.
+    1. Let |stream| be the result of [=create a new FileSystemWritableFileStream|creating a new FileSystemWritableFileStream=]
```
annevk, line 377: This really needs to happen from a task. Same below.

index.bs
annevk: Actually, this used to be synchronous but now it runs in parallel, so I think we need to change this. If something was already in parallel and makes this mistake I think it's fine to leave it for a follow-up.

index.bs
```diff
@@ -1249,8 +1275,11 @@ The <dfn method for=FileSystemSyncAccessHandle>close()</dfn> method steps are:
 
 1. If [=this=]'s [=[[state]]=] is "`closed`", return.
 1. Set [=this=]'s [=[[state]]=] to "`closed`".
+1. Set |lockReleased| to false.
+1. Let |onLockReleasedAlgorithm| be this step: Set |lockReleased| to true.
```
annevk, line 1279:
```suggestion
1. Let |onLockReleasedAlgorithm| be this step: set |lockReleased| to true.
```

index.bs
```diff
@@ -1249,8 +1275,11 @@ The <dfn method for=FileSystemSyncAccessHandle>close()</dfn> method steps are:
...
 1. [=file entry/lock/release|Release the lock=] on
-   [=this=]'s [=FileSystemSyncAccessHandle/[[file]]=].
+   [=this=]'s [=FileSystemSyncAccessHandle/[[file]]=] with |onLockReleasedAlgorithm|.
+1. [=Pause=] until |lockReleased| is set to true.
```
annevk, line 1282:
```suggestion
1. [=Pause=] until |lockReleased| is true.
```

annevk (commented):
With some more task queueing I think this can land.

index.bs
a-sully: > And yeah, if you already have a task you're not in parallel and you can resolve promises as you wish. (Though sometimes you might still want to queue a task, it really depends on what kind of ordering you are after.)

Awesome, thanks for the clarification!

> Actually, this used to be synchronous but now it runs in parallel, so I think we need to change this. If something was already in parallel and makes this mistake I think it's fine to leave it for a follow-up.

SGTM. Let me know if the pattern I sketched out in https://github.com/whatwg/fs/pull/9#discussion_r1107406159 is appropriate here

index.bs
a-sully: Ack. I'll do this in a follow-up

index.bs
a-sully: This now runs in the `file system queue`

index.bs

index.bs
a-sully: Also now happens on a task on the `file system queue`

a-sully:
Updated to post all file system operations to a "file system queue", as we discussed offline. Lock take/release is once again sync (and must be called from that queue). Please feel free to nit-pick this as much as you please, since this pattern will be applied to all the other methods. Thanks!

index.bs
```diff
@@ -90,38 +90,47 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
+To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of
+"`exclusive`" or "`shared`" on a given [=file entry=] |file|:
 
 1. Let |lock| be the |file|'s [=file entry/lock=].
 1. Let |count| be the |file|'s [=file entry/shared lock count=].
 1. If |value| is "`exclusive`":
    1. If |lock| is "`open`":
      1. Set lock to "`taken-exclusive`".
-     1. Return true.
+     1. Return a value of "`success`".
```
annevk, line 106:
```suggestion
     1. Return "`success`".
```

index.bs
```diff
@@ -90,38 +90,47 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
 1. If |value| is "`shared`":
    1. If |lock| is "`open`":
      1. Set |lock| to "`taken-shared`".
      1. Set |count| to 1.
-     1. Return true.
+     1. Return a value of "`success`".
```
annevk, line 111:
```suggestion
     1. Return "`success`".
```

index.bs
```diff
@@ -90,38 +90,47 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
    1. Otherwise, if |lock| is "`taken-shared`":
-     1. Increase |count| by one.
-     1. Return true.
-1. Return false.
+     1. Increase |count| by 1.
+     1. Return a value of "`success`".
+1. Return a value of "`failure`".
```
annevk, line 115:
```suggestion
     1. Return "`success`".
1. Return "`failure`".
```

index.bs
```diff
@@ -90,38 +90,47 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
+
+Note: These steps must be run on the [=file system queue=].
```
annevk, line 117: You cannot require things in notes. And here we just need a statement of fact.
```suggestion
Note: These steps have to be run on the [=file system queue=].
```

index.bs
```diff
@@ -90,38 +90,47 @@ A <dfn export id=file>file entry</dfn> additionally consists of
...
 
 1. Let |lock| be the |file|'s associated [=file entry/lock=].
 1. Let |count| be the |file|'s [=file entry/shared lock count=].
 1. If |lock| is "`taken-shared`":
-  1. Decrease |count| by one.
+  1. Decrease |count| by 1.
   1. If |count| is 0, set |lock| to "`open`".
 1. Otherwise, set |lock| to "`open`".
 
+Note: These steps must be run on the [=file system queue=].
```
annevk, line 132:
```suggestion
Note: These steps have to be run on the [=file system queue=].
```

annevk (commented):
The overall setup here seems sound from a quick skim. The main problem is that "reject", "resolve", and "create a new X" are operations that need to happen on the main thread. And as such you need to do the "queue a storage task" dance for them.

Note that you don't necessarily need to replace each individually. You might be able to restructure things a bit so do all the in parallel work first and safe the results in variables and then queue a storage task that does the rejecting/resolving as per the variables.

Also, what kind of exception can "request access" throw, out of curiosity? Since we get that exception while in parallel, forwarding it probably doesn't entirely work. If we could normalize that to it not returning "granted" that would be better.

a-sully:
> The overall setup here seems sound from a quick skim. The main problem is that "reject", "resolve", and "create a new X" are operations that need to happen on the main thread. And as such you need to do the "queue a storage task" dance for them.

Oh I see... Question - does this work? If not, why not?
```
1. [=Queue a storage task=] with [=this=]'s [=relevant global object=] to
   [=enqueue the following steps=] to the [=file system queue=]:
   1. ...
```

> Note that you don't necessarily need to replace each individually. You might be able to restructure things a bit so do all the in parallel work first and safe the results in variables and then queue a storage task that does the rejecting/resolving as per the variables.
> 
> Also, what kind of exception can "request access" throw, out of curiosity? Since we get that exception while in parallel, forwarding it probably doesn't entirely work. If we could normalize that to it not returning "granted" that would be better.

Unfortunately I think this is needed, since it can reject with a `SecurityError`. See https://wicg.github.io/file-system-access/#permissions

annevk:
> Question - does this work? If not, why not?

Because you still end up in parallel when doing "resolve", etc. You need some things to happen "in parallel" and then after those things you queue a task for things to happen on the main thread. Such as resolving a promise or creating a JS object.

> Unfortunately I think this is needed, since it can reject with a SecurityError.

Hmm, I guess we should redesign this somewhat whereby "request access" returns some kind of internal value and we then later turn that into the appropriate exception.

a-sully:
> > Question - does this work? If not, why not?
> 
> Because you still end up in parallel when doing "resolve", etc. You need some things to happen "in parallel" and then after those things you queue a task for things to happen on the main thread. Such as resolving a promise or creating a JS object.

Ah yes, looking at this with fresh eyes I see the problem. I've updated a few methods - if that pattern looks good, I'll copy it everywhere else

> > Unfortunately I think this is needed, since it can reject with a SecurityError.
> 
> Hmm, I guess we should redesign this somewhat whereby "request access" returns some kind of internal value and we then later turn that into the appropriate exception.

This is a bit tricky because the [WICG spec considers file system access](https://wicg.github.io/file-system-access/#dom-permissionname-file-system) a "[powerful feature](https://w3c.github.io/permissions/#dfn-powerful-feature)", and powerful features have [permission request algorithms](https://wicg.github.io/permissions-request/#permission-request-algorithm) that are allowed to throw. I've added an explicit warning that this spec doesn't comply with that pattern, which seems like it should be good enough?

Anyways, I've updated the access checks to return either a PermissionState or an error code that can be turned into a DomException (though I wonder if _returning_ (as opposed to throwing) a DomException directly is an allowed pattern?)

If this all seems reasonable, I'll file an issue to update the WICG spec (and update all the instances in this spec which assume these algorithms can throw)

index.bs
```diff
@@ -61,16 +61,31 @@ different storage mechanism with a different API for such files. The entry point
...
+listed in the [=error names table=].
+Unless specified otherwise it returns "{{PermissionState/denied}}".
+
+<p class=warning> Dependent specifications may consider this API a
+[=powerful feature=]. However, unlike other [=powerful features=] whose
+[=permission request algorithm=] may throw, [=/file system entry=]'s
+[=file system entry/query access=] and [=file system entry/request access=]
+algorithms must run [=in parallel=] on the [=file system queue=] and are
+therefore not allowed to throw. Instead, the caller is expected to [=/reject=]
+as appropriate should these algorithms return an [=exception/error name=].
```
a-sully, line 84: This really should go in WICG spec, but since we're doing something a bit funky it might be useful to have here, too?

index.bs
```diff
@@ -1032,29 +1079,43 @@ To
...
+        1. Set |stream|'s [=FileSystemWritableFileStream/[[file]]=]'s
+           [=file entry/binary data=] to |stream|'s [=[[buffer]]=].
+           If that throws an exception, [=/reject=] |closeResult| with that
+           exception and abort these steps.
+
+           Note: It is expected that this atomically updates the contents of the
+           file on disk being written to.
+
+        1. [=file entry/lock/release|Release the lock=] on
+           |stream|'s [=FileSystemWritableFileStream/[[file]]=].
```
a-sully, line 1111: Looking at this again - since we're running in a "storage task" at this point (albeit one posted from the "file system queue"), do we need to explicitly queue this to the "file system queue"?

a-sully:
> if that pattern looks good, I'll copy it everywhere else

friendly ping @annevk. This infra PR is holding up the PRs specifying `move()` and `remove()`

index.bs
annevk: If they run in parallel the caller will also be in parallel. They wouldn't be able to immediately reject something.

But also I'm not a big fan of this design. Perhaps these should return a struct?

index.bs
```diff
@@ -494,25 +518,35 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method
...
      [=file system entry/request access=] given "`readwrite`".
-     If that throws an exception, [=reject=] |result| with that exception and abort.
-  1. If |access| is not "{{PermissionState/granted}}",
-     [=reject=] |result| with a "{{NotAllowedError}}" {{DOMException}} and abort.
-
-  1. If |entry| is `null`, [=/reject=] |result| with a
-     "{{NotFoundError}}" {{DOMException}} and abort.
+  1. If |access| is not "{{PermissionState/granted}}":
+    1. Set |requestAccessError| to |access| if |access| is an
+       [=exception/error name=]; otherwise, "{{NotAllowedError}}".
```
annevk, line 529:
```suggestion
       [=exception/error name=]; otherwise "{{NotAllowedError}}".
```

index.bs
annevk: Yes. Because these steps run on the main thread, not the "file system queue" thread.

annevk (commented):
Thanks for your patience. This is looking rather good. Except the access checks, those could do with some more design work.

index.bs
a-sully: > If they run in parallel the caller will also be in parallel. They wouldn't be able to immediately reject something.

Updated to say `the caller is expected to [=queue a storage task=] to [=/reject=], as appropriate`

... though while going through the whole spec I realized that there is one situation - the _asynchronous iterator initialization steps_ - where there is no promise to reject. See the other comment

> But also I'm not a big fan of this design. Perhaps these should return a struct?

Done. AFAICT a struct item cannot be "optional" (only algorithm parameters?) so "error name" is just empty most of the time

index.bs
```diff
@@ -654,13 +730,17 @@ support for example recursive iteration.
...
+  1. Let |entry| be the result of [=locating an entry=]
+     given |handle|'s [=FileSystemHandle/locator=].
+  1. Let |accessResult| be the result of running |entry|'s
+     [=file system entry/query access=] given "`read`".
+  1. If |accessResult|'s [=file system access result/permission state=]
+     is not "{{PermissionState/granted}}":
+    1. Set |accessErrorName| to |accessResult|'s
+       [=file system access result/error name=] if it is not
+       « the empty string »; otherwise "{{NotAllowedError}}".
+    1. [=Throw=] an |accessErrorName| {{DOMException}}.
```
a-sully, line 743: We've specified that `query access` requires going in parallel on the `file system queue`, but if we get an exception there's no associated promise to reject...

... @mkruisselbrink do you know if this is this even necessary? Or is it enough to check `query access` in just the "get the next iteration result" algorithm below?

index.bs
```diff
@@ -671,43 +751,54 @@ To [=get the next iteration result=] for a {{FileSystemDirectoryHandle}} |handle
...
+    1. If |accessResult|'s [=file system access result/permission state=]
+       is not "{{PermissionState/granted}}":
+      1. Set |accessErrorName| to |accessResult|'s
+         [=file system access result/error name=] if it is not
+         « the empty string »; otherwise "{{NotAllowedError}}".
+      1. [=/Reject=] |promise| with an |accessErrorName| {{DOMException}}
+         and return |promise|.
+
+    1. If |directory| is `null`, [=/reject=] |result| with a
+       "{{NotFoundError}}" {{DOMException}} and abort these steps.
```
a-sully, line 771: Note: this is the only "behavioral" change in this PR. Presumably the "NotAllowedError" should take precedence over the "NotAllowedError", just like everywhere else

index.bs
```diff
@@ -135,7 +176,7 @@ A <dfn export id=directory>directory entry</dfn> additionally consists of a [=/s
 <dfn for="directory entry">children</dfn>, which are themselves [=/file system entries=].
 Each member is either a [=/file entry=] or a [=/directory entry=].
 
-A [=/file system entry=] |entry| should be [=list/contained=] in the [=children=] of at most one
+A [=/file system entry=] |entry| should be [=list/contained=] in the [=directory entry/children=] of at most one
```
a-sully, line 179: FYI: this fixes a spec build failure

a-sully (commented):
As promised, I've updated the whole spec to follow the pattern. Note that much of the diff is just formatting (e.g. adding an indentation)

I've manually double-checked that all promise rejections and resolutions now happen from the "storage task source" and all access checks are done in-parallel from the file system queue (though it's still possible I've missed something...)

index.bs
```diff
@@ -61,16 +61,48 @@ different storage mechanism with a different API for such files. The entry point
...
-
-Note: Implementations that only implement this specification and not dependent specifications do not
-need to bother implementing [=/file system entry=]'s [=file system entry/query access=] and [=file system entry/request access=].
+Each [=/file system entry=] has an associated
+<dfn for="file system entry" id=entry-query-access>query access</dfn>
+algorithm, which takes "`read`" or "`readwrite`" <var ignore>mode</var> and
+returns a [=/file system access result=].
+Unless specified otherwise it returns a [=/file system access result=] with a
+[=file system access result/permission state=] of "{{PermissionState/denied}}"
+and with a [=file system access result/error name=] of « the empty string ».
```
annevk, line 70: Wrapping this in « » is wrong as this is not a list. Applies quite a few times.

index.bs
```diff
@@ -61,16 +61,48 @@ different storage mechanism with a different API for such files. The entry point
...
+[=file system entry/request access|requesting=] access to the file system.
+It has the following [=struct/items=]:
+
+: <dfn for="file system access result">permission state</dfn>
+:: A {{PermissionState}}
+: <dfn for="file system access result">error name</dfn>
+:: A [=string=] which must be either « the empty string » or an
+   [=exception/error name=] listed in the [=error names table=].
+   If [=file system access result/permission state=] is
+   "{{PermissionState/granted}}" this must be « the empty string »
```
annevk, line 91: Can we make it so that when permission is denied it has to be an error name so we don't have to do defaulting below?

index.bs
annevk: It doesn't seem to do anything based on these steps. Per https://webidl.spec.whatwg.org/#idl-async-iterable we can omit it I think.

cc @domenic

index.bs
a-sully: brain fart. This is fixed in the latest patch

index.bs
a-sully: Done. This simplifies things quite a lot

Side note: is "iff" okay to use in specs? Happy to spell it out if not

index.bs
a-sully: Tentatively omitted it for now, pending input from others

a-sully:
@annevk friendly ping (this is holding up a number of other changes). If this looks good, feel free to merge

index.bs
```diff
@@ -61,16 +61,51 @@ different storage mechanism with a different API for such files. The entry point
...
-
-Note: Implementations that only implement this specification and not dependent specifications do not
-need to bother implementing [=/file system entry=]'s [=file system entry/query access=] and [=file system entry/request access=].
+Each [=/file system entry=] has an associated
+<dfn for="file system entry" id=entry-query-access>query access</dfn>
+algorithm, which takes "`read`" or "`readwrite`" <var ignore>mode</var> and
+returns a [=/file system access result=].
+Unless specified otherwise it returns a [=/file system access result=] with a
+[=file system access result/permission state=] of "{{PermissionState/denied}}"
+and with a [=file system access result/error name=] of the empty string.
```
annevk, line 70:
```suggestion
and with an [=file system access result/error name=] of the empty string.
```

index.bs
```diff
@@ -61,16 +61,51 @@ different storage mechanism with a different API for such files. The entry point
...
+
+Each [=/file system entry=] has an associated
+<dfn for="file system entry" id=entry-request-access>request access</dfn>
+algorithm, which takes "`read`" or "`readwrite`" <var ignore>mode</var> and
+returns a [=/file system access result=].
+Unless specified otherwise it returns a [=/file system access result=] with a
+[=file system access result/permission state=] of "{{PermissionState/denied}}"
+and with a [=file system access result/error name=] of the empty string.
```
annevk, line 78:
```suggestion
and with an [=file system access result/error name=] of the empty string.
```

annevk (commented):
Found a couple remaining nits, generally looks good. Feel free to merge after addressing them. Though I should have time tomorrow if you'd rather I do it.

annevk (approved):
I meant to approve this to make that possible. Doh.

index.bs
```diff
@@ -61,16 +61,51 @@ different storage mechanism with a different API for such files. The entry point
...
+result of [=file system entry/query access|querying=] or
+[=file system entry/request access|requesting=] access to the file system.
+It has the following [=struct/items=]:
+
+: <dfn for="file system access result">permission state</dfn>
+:: A {{PermissionState}}
+: <dfn for="file system access result">error name</dfn>
+:: A [=string=] which must be either the empty string or an
+   [=exception/error name=] listed in the [=error names table=].
+   Iff [=file system access result/permission state=] is
```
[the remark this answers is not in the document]
a-sully: Ack. Converted into a requirement (i.e. "will" -> "must" or "should")

a-sully:
Thanks for the review! I'll merge this now so we can start making progress on the various PRs that are blocked on it. If you have more nits, I'm happy to address them in a follow-up

Just an FYI that I also updated the permission checks to use the new `[=DOMException/name=]` and "names table" added in https://github.com/whatwg/webidl/pull/1211 (which broke the refs we were adding)

Issue #92, opened 2023-01-13
Consider renaming " the Origin Private File System" in anticipation of Storage Buckets


a-sully:
The Storage Buckets API (see https://wicg.github.io/storage-buckets/explainer) will allow sites to partition their storage into a number of buckets. Each storage bucket will store data associated with established storage APIs such as IndexedDB, CacheStorage, and, you guessed it, File System. Soon, a site may have multiple instances of "the Origin Private File System."

Should we change the name of the OPFS to reflect this? If so, does anyone have suggestions? Or does the OPFS name have enough behind it that changing it would be _more_ confusing to developers than leaving things as they are?

At the very least, we could change "**the** Origin Private File System" to "**an** Origin Private File System"? "Private" doesn't necessarily imply exclusivity... Sure, "Origin" is technically no longer correct, but maybe that distinction doesn't actually matter?

inexorabletash:
"Sandboxed" is the generic adjective that comes to mind. I'll see if I can dig up where we ended up with OPFS...

a-sully:
Yeah I think it may have at one point been referred to as the "sandboxed file system" in the spec, since the `3. Accessing the Origin Private File System` header has the following link: https://fs.spec.whatwg.org/#sandboxed-filesystem

inexorabletash:
Hmmm, https://github.com/WICG/file-system-access/pull/199 moved explicitly away from "sandboxed"...

mkruisselbrink:
https://github.com/WICG/file-system-access/pull/174#discussion_r410367052 is the comment where @inexorabletash suggested the current name instead of sandboxed.

mkruisselbrink:
Although to be fair, that was more a comment about the javascript API (which changed since then anyway) rather than the spec language.

inexorabletash:
Nice digging! And yeah, that explains why we avoided "Sandboxed" there - just wanting to not enshrine that term in the API itself without further discussion. So... no strong reason to _not_ use the term if we want it for describing the concept.

asutherland:
I would strongly suggest avoiding using the term "sandbox".  It has a very specific meaning in the HTML and CSP space that is distinct from OPFS' characteristics.  I believe there is also a browser vendor using the term "Privacy Sandbox" recently that also seems to be distinct from OPFS and could result in confusion.

annevk:
Perhaps "Local File System" to mimic `localStorage`?

asutherland:
[...]

Issue #80, opened 2022-12-21
Request to check if FileHandle/DirectoryHandle exists before executing read/write operations


poirierlouis:
After a small talk on Matrix chat (@annevk), I'd like to present a problem and use-case regarding handles when they no longer exists on user's device.

Currently, read/write access operations on `FileSystemFileHandle` and `FileSystemDirectoryHandle` are the only way to test if a file or directory entry exists on user's device. I've been explained there is no `exists()` like-method on these interfaces due to the `TOCTTOU` principle. Considering it, here is a problem where I believe current state of API prevent any chance at green code practice:

#### A. Use-case on file:
1. Get handle of a file (it can be from IndexedDB storage, a dialog picker...)
2. Prepare data (run long computation / algorithm, e.g. compress data, serialize data...)
3. Write data in file

Access to underlying file will happen at step (3), and when file does not exist, it rejects with a `NotFoundError`.
In such case, there is no way to save time and device's resources. It will run, depending on the use-case, a long computation / algorithm to prepare data. Only once data is ready, it will try to write in file, which may or may not exist.

#### B. Use-case on directory:
1. Get handle of a directory
2. Prepare structure of a project
    - list directories
    - list files
    - prepare data of files
3. Create prepared directories in (1)
4. Create prepared files in (1)

In the same situation, step (2) could be prevented if directory entry is already known to *not* exist.

#### Problem:
There is no way nor "chance" given at the end-user to prevent execution of step (2). This seems to be a waste regarding the device power/resources, and regarding the time of the user of the application.

Let me know if anything is not clear.
What do you think?

*Aside:* is TOCTTOU like a general no-go rule when writing a specification? Or can it be argued on depending on use-case, the technicality of such, I think, edge-cases? While not recommended, for example [Node.js](https://nodejs.org/api/fs.html#fsaccesspath-mode-callback) expose an `exists()` like-method.

dslee414:
I can see the argument that `exists()` like method may be useful, even though it won't guarantee the existence of file/directory at the time of update, as you mentioned. For example, `exists()` check may be done periodically while preparing the data, especially if it takes a while / is computation-heavy.

This might be relevant to `getMetadata()` support requested in https://github.com/whatwg/fs/issues/12. Could null metadata mean that the file/directory does not exist? or would it make more sense to have `exists()` method separately?

[...]

Issue #57, opened 2022-10-03
Specify error mappings


a-sully:
This came up in https://github.com/web-platform-tests/wpt/commit/b20e8343ce5ff1eada39dc1487b8b28ed903fff6#r85726061

Ideally API errors map uniquely to DOMExceptions, which make them actionable by the site. For example, if `NoModificationAllowedError` always indicates that a file is locked and `NotAllowedError` always indicates the site needs to `requestPermission()`on the handle, the site can respond differently to these errors.

Currently, developers have to figure out this mapping on their own, or read the whole spec and infer this mapping. It would be nice to include a table mapping API errors to DOMExceptions that developers can use as a reference. Is there precedent for having this type of table in a spec?

That being said, it could be tricky to add a mapping that truly _guarantees_ an error corresponds to a given API error (especially outside of the OPFS, where there's a whole host of things that could happen on the underlying OS). At the very least, here's a list of the mappings I'm aware of as a starting point...

- `NotAllowedError`
  - No permission at the API level (i.e. need to `requestPermission()`)
- `NoModificationAllowedError`
  - Cannot acquire a lock to the handle
  - Handle lacks read or write permission at the OS level (for example, trying to get write permission on a read-only file). This should only be relevant outside of the OPFS
- `InvalidModificationError`
  - Attempting to overwrite a file via `move()`
- `QuotaExceededError`
  - Operation will exceed quota
- `TypeMismatchError`
  - The handle type (file/dir) does not match what's on disk
- `AbortError`
  - Using a not-yet-launched feature (i.e. directory `move()`)
  - Some internal errors
- `InvalidStateError`
  - Attempting to close an already-closed writable or SyncAccessHandle
  - Some internal errors
 
Some errors are mostly relevant outside of the OPFS:

- `AbortError` (in addition to above)
  - Picker selection cancelled
  - File save/move blocked by Safe Browsing
 - `InvalidStateError` (in addition to above)
   - Trying to show a picker in a detached frame
- `SecurityError`
  - Trying to show picker in sandboxed iframe/opaque origin/third party context/without user gesture

annevk:
"AbortError" is very specific and mainly for abort signals only. cc @domenic 

I'd also recommend considering that implementations might want some flexibility going forward so requiring rather unique exceptions might not be worthwhile if web developers aren't expected to branch on the exception names anyway.

jesup:
[...]

Issue #11, opened 2022-03-07
Better specify promise rejection behavior


a-sully:
_Migrated from https://github.com/WICG/file-system-access/issues/68_ (see #2)

For example, the `getFileHandle()` and `getDirectoryHandle()` methods can reject, but how and with what exception?

PR #96, opened 2023-02-24
Give FileSystemHandles an associated path

### LINKED ISSUE #59: What is a FileSystemHandle?
[TPAC](https://github.com/whatwg/fs/issues/45#issuecomment-1248497482) highlighted a pretty substantial architectural difference between the Chromium and Firefox implementations of this API. Chromium's implementation maps a `FileSystemHandle` to a file path (which I had [attempted to codify a while back](https://github.com/whatwg/fs/pull/30)), while Firefox's implementation suggests "[they are references to objects resolved at creation time](https://github.com/whatwg/fs/issues/39#issuecomment-1248670783)."

This has gone unnoticed until this point because no features have exposed this difference to the web.

But there are some significant web-observable implications to this choice, most notably around directory moves. Consider the following code, which moves a directory containing a child we have a handle to:

```
// Create file at /old/file.txt.
const dir_entry = root.getDirectoryHandle('old', { create: true });
const file_entry = dir_entry.getFileHandle('file.txt', { create: true });

// The file previously pointed to file at /old/file.txt now resides at /new/file.txt.
await dir_entry.move('new');
```

What happens next?

```
// PATH-BASED: The promise rejects with NotFoundError.
// REF-BASED:  The call succeeds and returns the File.
await file_entry.getFile();

// PATH-BASED: The promise rejects with NotFoundError.
// REF-BASED:  The promise resolves successfully and creates a new writable stream.
await file_entry.createWritable();

// PATH-BASED: The promise resolves to `null`.
// REF-BASED:  The promise resolves to 'file.txt' (same as before the move).
await dir_entry.resolve(file_entry);

// PATH-BASED: The isSameEntry() promise resolves to `false`.
// REF-BASED:  The isSameEntry() promise resolves to `true`.
const other_file = await dir_entry.getFileHandle('file.txt');
await other_file.isSameEntry(file_entry);

// PATH-BASED: The directory's ID has changed, along with its path.
// REF-BASED:  The directory's ID remains unchanged since before the move.
await dir_entry.getUniqueId();
// PATH-BASED: The file's ID remains unchanged, along with its path.
// REF-BASED:  The file's ID remains unchanged since before the move.
await file_entry.getUniqueId();

// This method is proposed in https://github.com/whatwg/fs/issues/38
// PATH-BASED: The promise presumably will resolve to `null`.
// REF-BASED:  The promise presumably will resolve to a copy of `dir_entry`.
await file_entry.getParent();

// PATH-BASED: getFile() is once again valid and returns the file at /old/file.txt.
//             The isSameEntry() promise resolves to `true`.
// REF-BASED:  getFile() returns the file at /new/file.txt.
//             The isSameEntry() promise resolves to `false`.
const old_dir = await root.getDirectoryHandle('old', { create: true });
const other_file = await old_dir.getFileHandle('file.txt', { create: true });
await file_entry.getFile();
await other_file.isSameEntry(file_entry);
```

Directory moves _very_ blatantly expose this difference. It will be basically impossible to [specify directory moves](https://github.com/whatwg/fs/pull/10) in a way that's consistent across platforms without being much more specific here. This also has implications for at least [the `getUniqueId()`method](https://github.com/whatwg/fs/pull/46#issuecomment-1267411719) and [removed handles](https://github.com/whatwg/fs/issues/39#issuecomment-1248670783). Directory moves would also expose a difference in `resolve()` and a [discussed-but-not-yet-formally-proposed `getParent()` method](https://github.com/whatwg/fs/issues/38), among other things.

Looking at the code above, I tend to agree that a ref-based approach leads to outcomes which are more likely to align with user expectations, at least regarding directory moves. It would be nice if moving a directory didn't invalidate all child handles, for example. Meanwhile, there are some instances where a path-based approach arguably makes more sense. What happens to a `FileSystemHandle` when [its underlying file is removed](https://github.com/whatwg/fs/issues/39#issuecomment-1248670783)?

```
​​// Create a file.
const file_entry = root.getFileHandle('file.txt', { create: true });

// Remove the file.
await file_entry.remove();

// PATH-BASED: We can still write to a removed file.
// REF-BASED:  ????
await file_entry.createWritable();

// PATH-BASED: The isSameEntry() promise resolves to `true`.
// REF-BASED:  The isSameEntry() promise resolves to `false`, presumably?
const other_file = root.getFileHandle('file.txt', { create: true });
await other_file.isSameEntry(file_entry);
```

That being said, it seems reasonable to specify that a handle can still be written to if it's removed via this API, while handles removed by other means (such as by clearing site data or via an OS file manager) could be invalidated.

In summary...

One option is to never specify features that expose this implementation difference, such as directory moves. Unfortunately this is a pretty fundamental difference which I suspect will be hard to paper over as the API continues to evolve. To me, this is a very unsatisfying option. Consider a web IDE which just wants to rename `src/foo/` to `src/bar/`, but is forced to recursively copy all contained files. 

Another theoretical option is to accept that there will be cross-browser differences. Sure, moving a directory will invalidate all child handles, but you can re-acquire all the handles within the new directory. However, going down this path is bound to expose many more subtle cross-browser differences. For example, Chromium [locks all ancestor directories](https://crsrc.org/c/content/browser/file_system_access/file_system_access_write_lock_manager.cc;l=105) when a file has an open writable to ensure its path does not change while it is being written to. Having an open writable will block moving a parent directory, which would be a confusing restriction in a reference-based design. This option would be bad for developers and the impacted users, but just listing it here for completeness.

The other option is to specify a requirement that a `FileSystemHande` is a reference (in the same way I had attempted to specify it’s a path [here](https://github.com/whatwg/fs/pull/30)). That framework could be used to specify new methods such as `move()`, `remove()` and `getUniqueId()` in a way that’s consistent across browsers. This is our preferred option, but supporting this in Chromium would require a pretty substantial re-architecture that I'm hesitant to commit to without clear indication from other browsers and the developer community that handles should be based on references rather than paths...

@szewai does WebKit have an opinion here?

a-sully:
Fixes #59

A much more thorough (than #30) attempt at expressing that FileSystemHandles map to file paths. At a high level, it proposes that:

- A FileSystemHandle has an associated "file system locator"
- A "file system locator" contains, among other things, a "file system path"
- A "file system entry" can be located given a "file system locator"
- You can get a "file system locator" from a "file system entry"

See #59 for the implications of this in practice.

   * Chromium
   * WebKit
   * Specifying that FileSystemHandles are path-based will allow us to clarify many tests which are currently intentionally vague  as to hide the ambiguity whether handles are paths or references
   * Chromium: https://crbug.com/1426855
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1823961
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=254276

***
<a href="https://whatpr.org/fs/96.html" title="Last updated on Mar 22, 2023, 5:32 PM UTC (44897f1)">Preview</a> | <a href="https://whatpr.org/fs/96/bb07d59...44897f1.html" title="Last updated on Mar 22, 2023, 5:32 PM UTC (44897f1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -169,7 +169,7 @@ To <dfn for="file system locator" id=locator-resolve>resolve</dfn> a
   1. Let |childPath| be |child|'s [=FileSystemHandle/locator=]'s [=file system locator/path=].
   1. Let |rootPath| be |root|'s [=FileSystemHandle/locator=]'s [=file system locator/path=].
   1. If |childPath| is [=the same path as=] |rootPath|,
-     [=/resolve=] |result| with an empty list, and abort.
+     [=/resolve=] |result| with « », and abort.
 
   1. If |rootPath|'s [=list/size=] is greater than |childPath|'s [=list/size=],
      [=/resolve=] |result| with null, and abort.
@@ -229,11 +229,11 @@ these constraints:
 
 - If |locator| is a [=file locator=], they return a [=file entry=] or null.
 - If |locator| is a [=directory locator=], they return a [=directory entry=] or null.
-- If these steps return |entry|, then:
+- If these steps return a non-null |entry|, then:
   - [=Getting the locator=] with |entry| returns |locator|,
     provided no intermediate file system operations were run.
-  - |entry|'s [=file system entry/name=] must be the same [=string=] as
-    the last [=list/item=] of |locator|'s [=file system locator/path=].
+  - |entry|'s [=file system entry/name=] is the last [=list/item=] of |locator|'s
+    [=file system locator/path=].
 
 </div>
 
@@ -247,8 +247,8 @@ constraints:
 - If these steps return |locator|, then:
   - [=Locating an entry=] with |locator| returns |entry|,
     provided no intermediate file system operations were run.
-  - |entry|'s [=file system entry/name=] must be the same [=string=] as
-    the last [=list/item=] of |locator|'s [=file system locator/path=].
+  - |entry|'s [=file system entry/name=] is the last [=list/item=] of |locator|'s
+    [=file system locator/path=].
 
 </div>
 
@@ -291,9 +291,8 @@ interface FileSystemHandle {
 };
 </xmp>
 
-A {{FileSystemHandle}} object represents a [=/file system locator=].
-Each {{FileSystemHandle}} object is associated with a
-<dfn for=FileSystemHandle export>locator</dfn> (a [=/file system locator=]).
+A {{FileSystemHandle}} object is associated with a <dfn for=FileSystemHandle export>locator</dfn> (a
+[=/file system locator=]).
 
 Note: Multiple {{FileSystemHandle}} objects can have
 [=the same locator as|the same=] [=/file system locator=].
--- a/index.bs
+++ b/index.bs
@@ -60,8 +60,6 @@ different storage mechanism with a different API for such files. The entry point
 
 A <dfn export id="entry">file system entry</dfn> is either a [=file entry=] or a [=directory entry=].
 
-<!-- TODO: Consider tying access checks to a path, to support handles which don't map to an entry -->
-
 Each [=/file system entry=] has an associated <dfn for="file system entry" id=entry-query-access>query access</dfn> algorithm, which takes "`read`"
 or "`readwrite`" <var ignore>mode</var> and returns a {{PermissionState}}. Unless specified
 otherwise it returns "{{PermissionState/denied}}". The algorithm is allowed to throw.
@@ -92,8 +90,6 @@ A <dfn export id=file>file entry</dfn> additionally consists of
 <dfn for="file entry">modification timestamp</dfn> (a number representing the number of milliseconds since the <a spec=FileAPI>Unix Epoch</a>),
 a <dfn for="file entry">lock</dfn> (a string that may exclusively be "`open`", "`taken-exclusive`" or "`taken-shared`")
 and a <dfn for="file entry">shared lock count</dfn> (a number representing the number shared locks that are taken at a given point in time).
-
-<!-- TODO: Consider tying a locks to a path, to support moves -->
 
 <div algorithm>
 To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of "`exclusive`" or "`shared`" on a given [=file entry=] |file|:
@@ -153,8 +149,8 @@ as well as how changes made to the data structures defined here are reflected ex
 is left up to individual user-agent implementations.
 
 A [=/file system entry=] |a| is <dfn for="file system entry">the same entry as</dfn>
-a [=/file system entry=] |b| if |a| is equal to |b|, or if |a| and |b| are
-backed by the same file or directory on the local file system.
+a [=/file system entry=] |b| if |a| is equal to |b|, or if [=getting the path=]
+of |a| returns [=the same path as=] [=getting the path=] of |b|.
 
 <div algorithm>
 
@@ -174,37 +170,45 @@ run these steps:
         1. [=list/Prepend=] |entry|'s [=file system entry/name=] to |path|.
         1. [=/Resolve=] |result| with |path|.
   1. [=Wait for all=] |childPromises|, with the these success steps:
-    1. If |result| hasn't been resolved yet, [=/resolve=] |result| with `null`.
+    1. If |result| hasn't been resolved yet, [=/resolve=] |result| with null.
 1. Return |result|.
 
 </div>
 
-A <dfn export>file system path</dfn> represents a potential location of a
-[=/file system entry=]. For example, this may be to a virtual path that is
-mapped to real location on disk or in memory, may correspond directly to a
-path on the local file system, or may not correspond to any file on disk at all.
+A <dfn export>file system path</dfn> is an opaque string that represents a
+potential location of a [=/file system entry=].
+For example, this may be a virtual path that is mapped to real location on disk
+or in memory, may correspond directly to a path on the local file system, or may
+not correspond to any file on disk at all.
 
 The <dfn local-lt=locating>locate</dfn> algorithm takes a [=/file system path=] and
-returns either a [=/file system entry=] or `null`. The steps of this algorithm
+returns either a [=/file system entry=] or null. The steps of this algorithm
 are [=implementation-defined=].
 
 The <dfn local-lt="getting the path">get the path</dfn> algorithm takes a
 [=/file system entry=] and returns a [=/file system path=].
-If [=locating=] a [=/file system path=] |path| returns a [=/file system entry=] |entry|,
-then [=getting the path=] of |entry| must return |path|.
+If [=locating=] a [=/file system path=] |path| returns a [=/file system entry=]
+|entry|, then [=getting the path=] of |entry| must return |path| if no
+intermediate file system operations were run. Typically the steps are symmetric
+to the [=locate=] algorithm.
 
-Each [=/file system path=] has an associated <dfn for="file system path">name</dfn> (a [=string=]),
-which represents the last path component of the file path. For example, if
-[=/file system path=] |path| corresponds to a location represented by a
-virtual path of `data/drafts/example.txt` relative to the root directory of an
-[=origin private file system=], |path|'s [=file system path/name=] should be `example.txt`.
+Each [=/file system path=] has an associated <dfn for="file system path">name</dfn> (a [=string=]).
+If [=locating=] [=/file system path=] |path| returns a
+[=/file system entry=] |entry|, then |entry|'s [=file system entry/name=]
+must be the same as |path|'s[=file system path/name=].
+
+Note: This generally represents the last path component of the file path as
+perceived by the user. For example, if [=/file system path=] |path| corresponds
+to a location represented by a virtual path of `data/drafts/example.txt`
+relative to the root directory of an [=origin private file system=], |path|'s
+[=file system path/name=] should be `example.txt`. There is no expectation,
+however, that a file called `example.txt` lives anywhere on disk.
 
 If [=locating=] a [=/file system path=] |path| returns a [=/file system entry=] |entry|, then
 |path|'s [=file system path/name=] must equal |entry|'s [=file system entry/name=].
 
 A [=/file system path=] |a| is <dfn for="file system path">the same path as</dfn>
-a [=/file system path=] |b| if |a| represents the same potential location of a
-[=/file system entry=] as |b|.
+a [=/file system path=] |b| if |a| is |b|.
 
 ## The {{FileSystemHandle}} interface ## {#api-filesystemhandle}
 
@@ -226,8 +230,9 @@ interface FileSystemHandle {
 A {{FileSystemHandle}} object represents a [=/file system path=].
 Each {{FileSystemHandle}} object is associated with a
 <dfn for=FileSystemHandle export>path</dfn> (a [=/file system path=]).
-Multiple separate objects implementing the {{FileSystemHandle}} interface can
-all be associated with the same [=/file system path=] simultaneously.
+
+Note: Multiple {{FileSystemHandle}} objects can have
+[=the same path as|the same=] [=path=].
 
 <div algorithm>
 To <dfn for=FileSystemHandle id=get-handle-entry local-lt="getting the entry">get the entry</dfn>
--- a/index.bs
+++ b/index.bs
@@ -19,6 +19,7 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
   type: dfn; text: realm; url: realm
 urlPrefix: https://storage.spec.whatwg.org/; spec: storage
   type: dfn; text: storage; url: site-storage
+  type: dfn; text: storage bucket; url: storage-bucket
 </pre>
 
 <style>
@@ -158,11 +159,28 @@ run these steps:
 
 1. Let |result| be [=a new promise=].
 1. Run these steps [=in parallel=]:
-  1. If |child|'s [=FileSystemHandle/locator=] is [=the same locator as=]
-     |root|'s [=FileSystemHandle/locator=],
+  1. If |child|'s [=FileSystemHandle/locator=]'s [=file system locator/root=]
+     is not |root|'s [=FileSystemHandle/locator=]'s [=file system locator/root=],
+     [=/resolve=] |result| with null, and abort.
+
+  1. Let |childPath| be |child|'s [=FileSystemHandle/locator=]'s [=file system locator/path=].
+  1. Let |rootPath| be |root|'s [=FileSystemHandle/locator=]'s [=file system locator/path=].
+  1. If |childPath| is [=the same path as=] |rootPath|,
      [=/resolve=] |result| with an empty list, and abort.
 
-   <!-- TODO(a-sully): Figure out how to resolve locators. -->
+  1. If |rootPath|'s [=list/size=] is greater than |childPath|'s [=list/size=],
+     [=/resolve=] |result| with null, and abort.
+
+  1. [=list/For each=] |index| of |rootPath|'s [=list/indices=]:
+    1. If |rootPath|.\[[|index|]] is not |childPath|.\[[|index|]], then
+       [=/resolve=] |result| with null, and abort.
+
+  1. Let |relativePath| be an empty [=/list=].
+  1. [=list/For each=] |index| of [=the range=] from |rootPath|'s [=list/size=]
+     to |rootPath|'s [=list/size=], exclusive,
+     [=list/append=] |childPath|.\[[|index|]] to |relativePath|.
+
+  1. [=/Resolve=] |result| with |relativePath|.
 
 1. Return |result|.
 
@@ -172,24 +190,40 @@ A <dfn export>file system locator</dfn> represents a potential location of a
 [=/file system entry=]. A [=/file system locator=] is either a [=file locator=]
 or a [=directory locator=].
 
-Each [=/file system locator=] has an associated <dfn for="file system locator" id=locator-path>path</dfn> (a [=/file system path=])
-and an associated <dfn for="file system locator" id=locator-kind>kind</dfn> (a {{FileSystemHandleKind}}).
+Each [=/file system locator=] has an associated <dfn export for="file system locator" id=locator-path>path</dfn> (a [=/file system path=]),
+a <dfn export for="file system locator" id=locator-kind>kind</dfn> (a {{FileSystemHandleKind}}), and
+a <dfn export for="file system locator" id=locator-root>root</dfn> (an opaque string).
 
-A <dfn>file locator</dfn> is a [=/file system locator=] whose
+A <dfn export>file locator</dfn> is a [=/file system locator=] whose
 [=file system locator/kind=] is {{FileSystemHandleKind/"file"}}.
-A <dfn>directory locator</dfn> is a [=/file system locator=] whose
+A <dfn export>directory locator</dfn> is a [=/file system locator=] whose
 [=file system locator/kind=] is {{FileSystemHandleKind/"directory"}}.
 
-A [=/file system locator=] |a| is <dfn for="file system locator">the same locator as</dfn>
-a [=/file system locator=] |b| if |a| is |b|.
+A <dfn export>file system root</dfn> is an opaque string whose value is
+[=implementation-defined=].
 
-The <dfn local-lt="locating an entry">locate an entry</dfn> algorithm takes a
+<p class=example id=example-locator>For a [=/file system locator=] |locator|
+whichs [=locate an entry|locates to=] a [=file entry=] |entry| that conceptually
+exists at the path `data/drafts/example.txt` relative to the root directory of
+the [=origin private file system=],
+|locator|'s [=file system locator/kind=] has to be {{FileSystemHandleKind/"file"}},
+|locator|'s [=file system locator/path=] might be « "`data`", "`drafts`", "`example.txt`" », and
+|locator|'s [=file system locator/root=] might include relevant identifying
+information such as the [=storage bucket=] and the disk drive.
+
+A [=/file system locator=] |a| is <dfn for="file system locator">the same locator as</dfn>
+a [=/file system locator=] |b| if
+|a|'s [=file system locator/kind=] is |b|'s [=file system locator/kind=],
+|a|'s [=file system locator/root=] is |b|'s [=file system locator/root=], and
+|a|'s [=file system locator/path=] is [=the same path as=] |b|'s [=file system locator/path=].
+
+The <dfn export data-lt="locating an entry">locate an entry</dfn> algorithm takes a
 [=/file system locator=] |locator| and returns either a [=/file system entry=] or null.
 If |locator| is a [=file locator=], this method must return either a [=file entry=] or null.
 If |locator| is a [=directory locator=], this method must return either a [=directory entry=] or null.
 The steps of this algorithm are [=implementation-defined=].
 
-The <dfn local-lt="getting the locator">get the locator</dfn> algorithm takes a
+The <dfn export data-lt="getting the locator">get the locator</dfn> algorithm takes a
 [=/file system entry=] |entry| and returns a [=/file system locator=] |locator|.
 
 If |entry| is a [=file entry=], |locator| must be a [=file locator=].
@@ -201,11 +235,10 @@ Typically the steps are symmetric to the [=locate an entry=] algorithm.
 
 If [=locating an entry=] given [=/file system locator=] |locator| returns a
 [=/file system entry=] |entry|, then |entry|'s [=file system entry/name=] must
-be the same [=valid file name=] as the last [=list/item=] of
+be the same [=string=] as the last [=list/item=] of
 |locator|'s [=file system locator/path=].
 
-A <dfn export>file system path</dfn> is a [=/list=] of one or more
-[=valid file name|valid file names=].
+A <dfn export>file system path</dfn> is a [=/list=] of one or more [=strings=].
 This may be a virtual path that is mapped to real location on disk or in memory,
 may correspond directly to a path on the local file system, or may not
 correspond to any file on disk at all. The actual physical location of the
@@ -214,6 +247,12 @@ corresponding [=/file system entry=] is [=implementation-defined=].
 <p class=example id=example-path>Let |path| be the [=/list=]
 « "`data`", "`drafts`", "`example.txt`" ».
 There is no expectation that a file named `example.txt` exists anywhere on disk.
+
+A [=/file system path=] |a| is <dfn for="file system path">the same path as</dfn>
+a [=/file system path=] |b| if
+|a|'s [=list/size=] is the same as |b|'s [=list/size=] and
+[=list/for each=] |index| of |a|'s [=list/indices=]
+|a|.\[[|index|]] is |b|.\[[|index|]].
 
 ## The {{FileSystemHandle}} interface ## {#api-filesystemhandle}
 
@@ -276,7 +315,7 @@ The <dfn attribute for=FileSystemHandle>kind</dfn> getter steps are to return
 [=this=]'s [=FileSystemHandle/locator=]'s [=file system locator/kind=].
 
 The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return
-the last [=list/item=] (a [=valid file name=]) of [=this=]'s
+the last [=list/item=] (a [=string=]) of [=this=]'s
 [=FileSystemHandle/locator=]'s [=file system locator/path=].
 
 ### The {{FileSystemHandle/isSameEntry()}} method ### {#api-filesystemhandle-issameentry}
--- a/index.bs
+++ b/index.bs
@@ -194,6 +194,8 @@ A <dfn export>file system locator</dfn> represents a potential location of a
 [=/file system entry=]. A [=/file system locator=] is either a [=file locator=]
 or a [=directory locator=].
 
+<!-- TODO(a-sully): Consider giving each locator a Storage Bucket. -->
+
 Each [=/file system locator=] has an associated <dfn export for="file system locator" id=locator-path>path</dfn> (a [=/file system path=]),
 a <dfn export for="file system locator" id=locator-kind>kind</dfn> (a {{FileSystemHandleKind}}), and
 a <dfn export for="file system locator" id=locator-root>root</dfn> (an opaque [=string=]).
@@ -264,7 +266,7 @@ a [=/file system path=] |b| if
 [=list/for each=] |index| of |a|'s [=list/indices=]
 |a|.\[[|index|]] is |b|.\[[|index|]].
 
-Note: The contents of a [=/file system locator=], including its
+<p class=warning> The contents of a [=/file system locator=], including its
 [=file system locator/path=], are not expected to be shared in their entirety
 with the website process. The [=/file system path=] might contain components
 which are not known to the website unless the [=/file system locator=] is later
@@ -332,7 +334,7 @@ The <dfn attribute for=FileSystemHandle>kind</dfn> getter steps are to return
 
 The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return
 the empty [=string=] if [=this=] is the root directory of the
-[=origin private file system=], otherwise the last [=list/item=] (a [=string=])
+[=origin private file system=]; otherwise the last [=list/item=] (a [=string=])
 of [=this=]'s [=FileSystemHandle/locator=]'s [=file system locator/path=].
 
 ### The {{FileSystemHandle/isSameEntry()}} method ### {#api-filesystemhandle-issameentry}
@@ -737,7 +739,7 @@ The <dfn method for=FileSystemDirectoryHandle>getFileHandle(|name|, |options|)</
 
 1. Let |result| be [=a new promise=].
 1. Let |realm| be [=this=]'s [=relevant Realm=].
-1. Let |locator| [=this=]'s [=FileSystemHandle/locator=].
+1. Let |locator| be [=this=]'s [=FileSystemHandle/locator=].
 1. Let |entry| be the result of [=locating an entry=] given |locator|.
 1. Run these steps [=in parallel=]:
   1. If |name| is not a [=valid file name=], [=/reject=] |result| with a {{TypeError}} and abort.
@@ -811,7 +813,7 @@ The <dfn method for=FileSystemDirectoryHandle>getDirectoryHandle(|name|, |option
 
 1. Let |result| be [=a new promise=].
 1. Let |realm| be [=this=]'s [=relevant Realm=].
-1. Let |locator| [=this=]'s [=FileSystemHandle/locator=].
+1. Let |locator| be [=this=]'s [=FileSystemHandle/locator=].
 1. Let |entry| be the result of [=locating an entry=] given |locator|.
 1. Run these steps [=in parallel=]:
   1. If |name| is not a [=valid file name=], [=/reject=] |result| with a {{TypeError}} and abort.
--- a/index.bs
+++ b/index.bs
@@ -61,8 +61,6 @@ different storage mechanism with a different API for such files. The entry point
 
 A <dfn export id="entry">file system entry</dfn> is either a [=file entry=] or a [=directory entry=].
 
-<!-- TODO(a-sully): Make these algorithms associated with a FileSystemHandle. -->
-
 Each [=/file system entry=] has an associated <dfn for="file system entry" id=entry-query-access>query access</dfn> algorithm, which takes "`read`"
 or "`readwrite`" <var ignore>mode</var> and returns a {{PermissionState}}. Unless specified
 otherwise it returns "{{PermissionState/denied}}". The algorithm is allowed to throw.
@@ -73,6 +71,8 @@ otherwise it returns "{{PermissionState/denied}}". The algorithm is allowed to t
 
 Note: Implementations that only implement this specification and not dependent specifications do not
 need to bother implementing [=/file system entry=]'s [=file system entry/query access=] and [=file system entry/request access=].
+
+Issue(101): Make access check algorithms associated with a FileSystemHandle.
 
 Each [=/file system entry=] has an associated <dfn for="file system entry" id=entry-name>name</dfn> (a [=string=]).
 
@@ -179,7 +179,7 @@ run these steps:
     1. If |rootPath|.\[[|index|]] is not |childPath|.\[[|index|]], then
        [=/resolve=] |result| with null, and abort.
 
-  1. Let |relativePath| be an empty [=/list=].
+  1. Let |relativePath| be « ».
   1. [=list/For each=] |index| of [=the range=] from |rootPath|'s [=list/size=]
      to |rootPath|'s [=list/size=], exclusive,
      [=list/append=] |childPath|.\[[|index|]] to |relativePath|.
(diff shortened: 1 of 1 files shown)


a-sully:
I expect the build to fail for now, since `[=FileSystemHandle/entry=]` has been un-defined. I plan to update all the uses once we decide on how a FileSystemHandle should relate to an entry

index.bs
```diff
@@ -176,6 +166,27 @@ run these steps:
...
+representing a potential location of a [=/file system entry=]. For example, this
+may be to a virtual path that is mapped to real location on disk or in memory,
+may correspond directly to a path on the local file system, or may not
+correspond to any file on disk at all.
+
+Each [=/file system path=] has an associated <dfn for="file system path">locate</dfn>
+algorithm, which returns a [=/file system entry=]. Exactly how a [=/file system path=]
+maps to a [=/file system entry=] left up to individual user-agent implementations.
+
+Each [=/file system path=] has an associated <dfn for="file system path" id=entry-query-access>query access</dfn> algorithm, which takes "`read`"
```
a-sully, line 179: `query access` and `request access` should map to a FileSystemHandle, and therefore a path, but I believe it's more semantically correct to have locks correspond to an entry (for example, POSIX's [`flock` takes a file descriptor](https://man7.org/linux/man-pages/man2/flock.2.html))

index.bs
```diff
@@ -193,9 +204,18 @@ interface FileSystemHandle {
 };
 </xmp>
 
-A {{FileSystemHandle}} object represents a [=/file system entry=]. Each {{FileSystemHandle}} object is associated
-with an <dfn for=FileSystemHandle export>entry</dfn> (a [=/file system entry=]). Multiple separate objects implementing
```
a-sully, line 197: Un-defining `[=FileSystemHandle/entry=]` breaks most of this spec (and https://github.com/WICG/file-system-access), and a FileSystemHandle _does_ conceptually have an associated entry... even if getting that entry requires looking it up using path.

Is it possible to have an "associated" member (i.e. `[=FileSystemHandle/entry=]`) that's retrieved using an algorithm?

It would be something like:
```
Each {{FileSystemHandle}} object is associated with a
<dfn for=FileSystemHandle export>path</dfn> (a [=/file system path=])
which maps to an associated <dfn for=FileSystemHandle export>entry</dfn> (a [=/file system entry=])
using an [=implementation-defined=] mapping.
```

index.bs
```diff
@@ -176,6 +166,27 @@ run these steps:
 
 </div>
 
+A <dfn export>file system path</dfn> is an implementation-defined type
+representing a potential location of a [=/file system entry=]. For example, this
+may be to a virtual path that is mapped to real location on disk or in memory,
+may correspond directly to a path on the local file system, or may not
+correspond to any file on disk at all.
```
a-sully, line 173: I'm happy to take suggestions on how to improve this definition, but it needs to be broad enough to allow sites to store as much information about the path/locator as needed, such as the context the path exists in (e.g. its bucket - see https://github.com/whatwg/fs/pull/30#issuecomment-1139068595)

index.bs
annevk: Isn't part of the problem that it doesn't always map to an entry? I think you'd have to change all call sites anyway to take that into account and given that it seems better to have a more explicit algorithm that takes you from one to the other.

index.bs
annevk: To allow sites or to allow user agents?

I think stating this as "A file system path represents the potential location of a file system entry." might be sufficient. As long as we're not directly exposing it, we don't have to define the details.

As for "locate", would it always return a file system entry or would it sometimes return nothing at all? E.g., if you need to create it first? I would probably make that a separate algorithm that takes a file system path. It seems a bit weird for something that I imagine would mostly be represented as a string or byte sequence to hold a concept such as locate.

a-sully: > To allow sites or to allow user agents?

Apologies, I meant "user agents"

> I think stating this as "A file system path represents the potential location of a file system entry." might be sufficient. As long as we're not directly exposing it, we don't have to define the details.
> 
> As for "locate", would it always return a file system entry or would it sometimes return nothing at all? E.g., if you need to create it first? I would probably make that a separate algorithm that takes a file system path. It seems a bit weird for something that I imagine would mostly be represented as a string or byte sequence to hold a concept such as locate.

Ack. I'll change the `[file system path/locate]` method to a "locate" function that takes a path

index.bs
a-sully: Yeah that's what I was thinking. It seems to me like the options are:

- a path can always map to an entry, which may "exist" or not (as currently sketched out in the `getFile()` method)
- the locate method returns either a valid entry, or nothing. And then perhaps we can add a `[=FileSystemHandle/exists=]` algorithm for simplicity. The latter seems more intuitive to me. I can test this out in the next patch...

index.bs
a-sully: That being said... for `move()`, we need to be able to lock the destination, where an entry does not yet exist. So what we're really locking is the potential location of a future entry, a.k.a. a path

index.bs
a-sully: > * the locate method returns either a valid entry, or nothing

This is sketched out in the latest patch!

index.bs
a-sully: Tentatively kept `query access` and `request access` as associated with an entry for now, but we might need to change this to support requesting access to handles which do not map to an entry

index.bs
```diff
@@ -90,6 +93,8 @@ A <dfn export id=file>file entry</dfn> additionally consists of
 a <dfn for="file entry">lock</dfn> (a string that may exclusively be "`open`", "`taken-exclusive`" or "`taken-shared`")
 and a <dfn for="file entry">shared lock count</dfn> (a number representing the number shared locks that are taken at a given point in time).
 
+<!-- TODO: Consider tying a locks to a path, to support moves -->
```
a-sully, line 96: Unless we're okay with locking entries which don't (yet) exist...

index.bs
```diff
@@ -59,6 +60,8 @@ different storage mechanism with a different API for such files. The entry point
 
 A <dfn export id="entry">file system entry</dfn> is either a [=file entry=] or a [=directory entry=].
 
+<!-- TODO: Consider tying access checks to a path, to support handles which don't map to an entry -->
```
mkruisselbrink, line 63: fwiw, at least originally we intentionally did not tie access checks to paths, to provide more flexibility for implementations (and I don't think even today our implementation actually ties access checks to path either?). One case where this makes a difference is:
- get a read-only directory handle via a directory picker
- get a writable file handle, that happens to be in the above directory
- now get a file handle to the same file by calling getFileHandle on the directory handle

I'm pretty sure that in at least our implementation the latter file handle will still be read-only. (i.e. while we sort of tie access checks to paths, they're not tied to the path of the entry a handle refers to, rather they are tied to the path of the file or directory that a user at some point picked).

index.bs
```diff
@@ -176,6 +179,33 @@ run these steps:
 
 </div>
 
+A <dfn export>file system path</dfn> represents a potential location of a
+[=/file system entry=]. For example, this may be to a virtual path that is
```
mkruisselbrink, line 183:
```suggestion
[=/file system entry=]. For example, this may be a virtual path that is
```

index.bs
```diff
@@ -176,6 +179,33 @@ run these steps:
...
+The <dfn local-lt="getting the path">get the path</dfn> algorithm takes a
+[=/file system entry=] and returns a [=/file system path=].
+If [=locating=] a [=/file system path=] |path| returns a [=/file system entry=] |entry|,
+then [=getting the path=] of |entry| must return |path|.
+
+Each [=/file system path=] has an associated <dfn for="file system path">name</dfn> (a [=string=]),
+which represents the last path component of the file path. For example, if
+[=/file system path=] |path| corresponds to a location represented by a
+virtual path of `data/drafts/example.txt` relative to the root directory of an
+[=origin private file system=], |path|'s [=file system path/name=] should be `example.txt`.
```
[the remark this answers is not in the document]
mkruisselbrink: Yeah, since path isn't exposed directly, do we even need this requirement? It seems you have two choices, either treat path as opaque (and not have this requirement), or define a lot more about the path structure.

Maybe what this is trying to say is that if two entries have the same parent and name they should also have the same path? (i.e. that names are unique?)

index.bs
mkruisselbrink: Related to my comment above on tying access checks to paths, the current chrome implementation (at least last I looked at it, which admittedly is a while ago, so maybe that has been changed already?) does not guarantee that two handles representing the same path behave the same in all aspects (since for handles acquired via DirectoryHandle.getFileHandle/getDirectoryHandle the access checks are done with the path of the original directory, rather than the path of the current handle).

I'm not saying that shouldn't be changed, but it's certainly something to keep in mind. Perhaps tying access checks to the FileSystemHandle rather than the entry would be a way around that, if we want to keep the same behavior? Although you'd have to figure out how to serialize that properly as well.

index.bs
a-sully: You're correct that, at least in Chrome's current implementation, the file handle you get from calling `getFileHandle()` on the directory handle is read-only, though it's debatable whether that's a feature or a bug. I can see this leading to confusion on the part of the website author and user. Regardless, that's a Chrome implementation detail...

The reason I mention this is because if we assert that an entry must exist on disk, then we can't use it to lock/check access to entries that don't exist, such as when locking the target file when renaming a file.

... thinking about this a bit more, perhaps the `move()` case isn't actually an issue if we specify that all file system operations to the same queue (#95), since that would guarantee (at least within the OPFS) that the target file isn't created and locked while the `move()` operation is in progress

Not sure what to do about access checks, though, if we assert that an "entry" exists on disk

index.bs
```diff
@@ -289,12 +341,18 @@ A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/entry=] must be a [=f
...
   1. Let |access| be the result of running [=this=]'s [=FileSystemHandle/entry=]'s
      [=file system entry/query access=] given "`read`".
   1. If |access| is not "{{PermissionState/granted}}",
      [=/reject=] |result| with a "{{NotAllowedError}}" {{DOMException}} and abort.
-  1. Let |entry| be [=this=]'s [=FileSystemHandle/entry=].
+
+  1. If |entry| is `null`, [=/reject=] |result| with a
+     "{{NotFoundError}}" {{DOMException}} and abort.
+  1. If |entry| is not a [=file entry=] [=/reject=] |result| with a
+     "{{TypeMismatchError}}" {{DOMException}} and abort.
```
a-sully, line 354: I realize (as per https://github.com/whatwg/fs/issues/57#issuecomment-1277022240) that `TypeMismatchError` is deprecated, but it's used elsewhere in this API and is thrown implicitly in a number of places already (e.g. most of the places affected by #11). This PR helps make it more explicit where this error comes from, but obviously it's not great to be throwing a deprecated exception so often... any suggestions on what to do here?

index.bs
```diff
@@ -271,7 +311,18 @@ interface FileSystemFileHandle : FileSystemHandle {
 };
 </xmp>
 
-A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/entry=] must be a [=file entry=].
+A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/path=]
+is expected to [=locate=] a [=file entry=].
```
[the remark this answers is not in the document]
a-sully: So, a `FileSystemDirectoryHandle` can also locate to a file entry. It just means that it shouldn't be used... for example:
```
let dirHandle = await root.getDirectoryHandle('test', { create: true });

await root.remove({ recursive: true });

let fileHandle = await root.getFileHandle('test', {create:true});
```
After these statements, `dirHandle` and `fileHandle` both correspond to the file "test", though operations (at least those which check the file system) on `dirHandle` will fail because the entry is the wrong type

I thought about overriding the "locate" algorithm for both `FileSystemDirectoryHandle` and `FileSystemFileHandle` to only return an entry if it's the correct type. Something like 
- let |result| be the result of locating the path
- return null if |result| is null
- return |result| if it's the correct type
- otherwise return "not a {file|directory}" 

And then `TypeMismatchError` can be thrown by the caller if it gets the "not a blah" string... but it seemed more clear (and provides a bit more flexibility) to just always check at the call site like:
```
  1. If |entry| is not a [=file entry=] [=/reject=] |result| with a
     "{{TypeMismatchError}}" {{DOMException}} and abort.
```

Thoughts?

index.bs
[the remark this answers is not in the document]
a-sully: > Maybe what this is trying to say is that if two entries have the same parent and name they should also have the same path? (i.e. that names are unique?)

I think that's already covered by the symmetry (assuming no intermediate operations are run, as Anne pointed out) between "locate" and "get the path" above

I can update this to explicitly mention that an entry's `name` should match the `name` of the path which locates it. WDYT?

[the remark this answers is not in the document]
a-sully: > Yeah, since path isn't exposed directly, do we even need this requirement? It seems you have two choices, either treat path as opaque (and not have this requirement), or define a lot more about the path structure.

Yeah it would be nice to keep the path opaque. My only concern is with the `resolve()` method. Consider the following scenario:
```
let root = await navigator.storage.getDirectory();
let dir = await root.getDirectoryHandle('test', {create:true});
let file = await dir.getFileHandle('example.txt', {create:true});

await dir.remove({ recursive: true });

// Currently returns ['test', 'example.txt'] on Chromium and Safari
await root.resolve(file); 
```

Locating `file` will return null, so `resolve()` as we've currently defined it doesn't really work. It seems like that method really should to take a path, as well?

index.bs
```diff
@@ -147,11 +152,9 @@ Exactly how external changes are reflected in the data structures defined by thi
...
 is left up to individual user-agent implementations.
 
-A [=/file system entry=] |a| is <dfn for="entry">the same as</dfn> an [=/file system entry=] |b| if |a| is equal to |b|, or
-if |a| and |b| are backed by the same file or directory on the local file system.
-
-Issue(59): Explain better how entries map to files on disk (multiple entries can map to the same file or
-directory on disk but an entry doesn't have to map to any file on disk).
+A [=/file system entry=] |a| is <dfn for="file system entry">the same entry as</dfn>
+a [=/file system entry=] |b| if |a| is equal to |b|, or if |a| and |b| are
+backed by the same file or directory on the local file system.
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -176,6 +179,33 @@ run these steps:
 
 </div>
 
+A <dfn export>file system path</dfn> represents a potential location of a
+[=/file system entry=]. For example, this may be to a virtual path that is
+mapped to real location on disk or in memory, may correspond directly to a
+path on the local file system, or may not correspond to any file on disk at all.
+
+The <dfn local-lt=locating>locate</dfn> algorithm takes a [=/file system path=] and
+returns either a [=/file system entry=] or `null`. The steps of this algorithm
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -176,6 +179,33 @@ run these steps:
...
+are [=implementation-defined=].
+
+The <dfn local-lt="getting the path">get the path</dfn> algorithm takes a
+[=/file system entry=] and returns a [=/file system path=].
+If [=locating=] a [=/file system path=] |path| returns a [=/file system entry=] |entry|,
+then [=getting the path=] of |entry| must return |path|.
```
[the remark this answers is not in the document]
a-sully: Added "if no intermediate file system operations were run". Is that precise enough?

index.bs
```diff
@@ -176,6 +179,33 @@ run these steps:
...
+[=/file system path=] |path| corresponds to a location represented by a
+virtual path of `data/drafts/example.txt` relative to the root directory of an
+[=origin private file system=], |path|'s [=file system path/name=] should be `example.txt`.
+
+If [=locating=] a [=/file system path=] |path| returns a [=/file system entry=] |entry|, then
+|path|'s [=file system path/name=] must equal |entry|'s [=file system entry/name=].
+
+A [=/file system path=] |a| is <dfn for="file system path">the same path as</dfn>
+a [=/file system path=] |b| if |a| represents the same potential location of a
+[=/file system entry=] as |b|.
```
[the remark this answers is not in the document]
a-sully: Done (for now - pending resolution of https://github.com/whatwg/fs/pull/96#discussion_r1122107537)

index.bs
```diff
@@ -193,17 +223,26 @@ interface FileSystemHandle {
...
 </xmp>
 
-A {{FileSystemHandle}} object represents a [=/file system entry=]. Each {{FileSystemHandle}} object is associated
-with an <dfn for=FileSystemHandle export>entry</dfn> (a [=/file system entry=]). Multiple separate objects implementing
-the {{FileSystemHandle}} interface can all be associated with the same [=/file system entry=] simultaneously.
+A {{FileSystemHandle}} object represents a [=/file system path=].
+Each {{FileSystemHandle}} object is associated with a
+<dfn for=FileSystemHandle export>path</dfn> (a [=/file system path=]).
+Multiple separate objects implementing the {{FileSystemHandle}} interface can
+all be associated with the same [=/file system path=] simultaneously.
```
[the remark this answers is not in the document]

index.bs
a-sully: good catch, thanks

a-sully:
@annevk do you have any idea why the build is failing? Does it not like "getting the entry"?

index.bs
a-sully: Sorry for my confusion, but why would we need to serialize access checks if we were to move them from the "entry" to the FileSystemHandle?

index.bs
a-sully: resolving this thread, since this is being discussed elsewhere

index.bs
a-sully: discussed in other threads. Posting all operations to the file system lock queue (see #95) would resolve this

index.bs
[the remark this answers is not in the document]
annevk: Yeah, that makes me think we should define a file system path as a list of strings, similar to how the URL Standard does it. We can still say that the actual physical location is implementation-defined.

annevk:
I suspect it's failing because of "the the". Search for "typos" in https://resources.whatwg.org/build/deploy.sh.

index.bs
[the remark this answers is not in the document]
annevk: If a path is a list of strings, how do we know it' a directory? Would the last string be the empty string? (That's how the URL Standard does it, although it doesn't really deal in directories per se.)

Because if we did that locating would either return null or the correct entry, right?

index.bs
annevk: I think aligning with @domenic's suggestions is still doable. Typically exceptions are not relied upon.

index.bs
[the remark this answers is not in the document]
a-sully: Is "_a_ is _b_" correct here if a path is a list of strings?

index.bs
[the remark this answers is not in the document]
a-sully: Implemented this suggestion. LMK what you think!

index.bs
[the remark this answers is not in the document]
a-sully: Is there an equivalent of "_a_ is _b_" for lists?

index.bs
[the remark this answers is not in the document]
a-sully: Added a "kind" to a "locator", which can now be a "file locator" or a "directory locator"

index.bs
a-sully: Since "locating an entry" now takes into account a "kind", I've changed this to assert that the "entry" you get back is of the right kind

a-sully:
> I suspect it's failing because of "the the". Search for "typos" in https://resources.whatwg.org/build/deploy.sh.

Ah, good catch. Thanks!

index.bs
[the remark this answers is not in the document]
annevk: There is not. You could check they are the same length. And then do a for each on https://infra.spec.whatwg.org/#list-get-the-indices and then do listA[index] is listB[index].

index.bs
```diff
@@ -147,35 +150,71 @@ Exactly how external changes are reflected in the data structures defined by thi
...
-    1. Let |p| be the result of [=file system entry/resolving=] |child| relative to |entry|.
-    1. [=list/Append=] |p| to |childPromises|.
-    1. [=Upon fulfillment=] of |p| with value |path|:
-      1. If |path| is not null:
-        1. [=list/Prepend=] |entry|'s [=file system entry/name=] to |path|.
-        1. [=/Resolve=] |result| with |path|.
-  1. [=Wait for all=] |childPromises|, with the these success steps:
-    1. If |result| hasn't been resolved yet, [=/resolve=] |result| with `null`.
+
+   <!-- TODO(a-sully): Figure out how to resolve locators. -->
```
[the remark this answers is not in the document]
annevk: Having an opaque root makes sense to me.

There's no standalone algorithm for this currently. The URL parser handles paths correctly.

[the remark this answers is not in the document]
a-sully: Added an opaque root and updated this algorithm!

index.bs
```diff
@@ -147,35 +151,109 @@ Exactly how external changes are reflected in the data structures defined by thi
...
+
+A <dfn export>file system root</dfn> is an opaque string whose value is
+[=implementation-defined=].
+
+<p class=example id=example-locator>For a [=/file system locator=] |locator|
+whichs [=locate an entry|locates to=] a [=file entry=] |entry| that conceptually
+exists at the path `data/drafts/example.txt` relative to the root directory of
+the [=origin private file system=],
+|locator|'s [=file system locator/kind=] has to be {{FileSystemHandleKind/"file"}},
+|locator|'s [=file system locator/path=] might be « "`data`", "`drafts`", "`example.txt`" », and
```
a-sully, line 210: I say "might" here because it's still "implementation defined" how a "locator" maps to an "entry", and most of the methods (with `isSameEntry()` and `resolve()` being the exceptions) require the "entry" to exist to do anything useful. For example, take the `getFileHandle(name)` method. It's basically:

1. Given a "locator", locate `entry` (a dir entry)
2. Given `entry`, find the `child` whose name is `name`
3. Create a FileSystemFileHandle given `child`, which involves "getting the locator" of `child`

So nowhere in this flow do we ever specify appending `name` to `this`'s locator's path to get `child`'s locator's path

Relatedly, creating the root of an OPFS doesn't say anything about its locator's path

If we want to assert that the path above "has to be" `/data/drafts/example.txt` (in list form) then we could:
- ensure that handles acquired from various API surfaces are given a defined path
  - set the path of the root of an OPFS to a list containing one empty string
  - outside of the OPFS (e.g. for a file picked from a file picker), let the path be "implementation defined"
- create handles from existing handles by manipulating the existing handle's locator's path
  - add new versions of the "create a new FileSystem{File|Directory}Handle" algorithms that append to the path of the existing handle

I'm happy to make this change (either here or in a follow-up), but this is where `getFileHandle()` and `getDirectoryHandle()` get thorny, as @szewai pointed out https://github.com/whatwg/fs/issues/59#issuecomment-1402897170). Arguably, these methods could ignore the state on disk and just always return a handle whose locator has had its path appended to, without bothering to check what exists on disk....

... but this would be a breaking change, and a pretty substantial one where `{ create: true }` is currently passed

For `{ create: false }` with `dir` that exists at path "/testing":
```
await dir.remove();
// Before: throws a NotFoundError
// After: succeeds - handle's path is "/testing/example.txt"
let handle = await dir.getFileHandle('example.txt', { create: false });
```
This is a web-exposed change, but arguably not that bad, especially if we add something like #80. Something that threw an exception before would longer throw.

For `{ create: true }` with `dir` that exists at path "/testing" if we deprecate the "create" option:
```
// Before: creates a new file at "/testing/example.txt"
// After: handle's path is "/testing/example.txt" - but no file is created
let handle = await dir.getFileHandle('example.txt', { create: true });

// At this point, any method which cares about what's on disk now fails

// Before: creates a SyncAccessHandle
// After: throws a NotFoundError
let accessHandle = await handle.createSyncAccessHandle();

// Before: gets the file's contents
// After: throws a NotFoundError
let file = await handle.getFile();
```

I worry that deprecating this option would break _a lot_ of existing code. Something that [works now](https://webkit.org/blog/12257/the-file-system-access-api-with-origin-private-file-system/#create:~:text=Creating%20a%20directory%20or%20a%20file) would consistently throw

My tentative thoughts are that we should just leave this as "might" in this PR, then make the changes above to turn this into a "has to be", then look into making `getFileHandle()` and `getDirectoryHandle()` (perhaps by adding an additional flag? We can hash that out later :) )

Thoughts?

annevk: Let me take a step back here to ensure I understand.

* `getFileHandle()` is exposed on `FileSystemDirectoryHandle`.
* A `FileSystemDirectoryHandle` is represented by a locator and as such when a method is invoked upon it that might fail if the locator no longer has a corresponding entry.

Why could we not append the given name to the locator's path and then see if an entry exists there? I think we can keep the existing behavior here, even if it's not necessarily the most natural given the "new" underpinnings.

---

I think for the OPFS root the path would indeed have to be the empty list. I'm not sure about making it contain an empty string. That only seems sensible if we did directories that way and you decided we wouldn't do that.

---

For file pickers you'd want the root to be a unique identifier and the path to be a list containing the file name, I think. At least by default `<input type=file>` shouldn't start exposing more about the user's machine.

a-sully: > Let me take a step back here to ensure I understand.
> 
> * `getFileHandle()` is exposed on `FileSystemDirectoryHandle`.
> * A `FileSystemDirectoryHandle` is represented by a locator and as such when a method is invoked upon it that might fail if the locator no longer has a corresponding entry.

Correct 👍 

> Why could we not append the given name to the locator's path and then see if an entry exists there? I think we can keep the existing behavior here, even if it's not necessarily the most natural given the "new" underpinnings.

That's basically what I was proposing on the other thread, so yeah that SGTM. I just wanted to explain what I understood of the other perspectives that were expressed (@szewai please correct me if I've misunderstood anything!)

The only nitpick here is that if `{ create: true }` currently fails rather than recursively create an entry. So we'd still need to check `this`'s locator first before appending the given name, or change the behavior

> I think for the OPFS root the path would indeed have to be the empty list. I'm not sure about making it contain an empty string. That only seems sensible if we did directories that way and you decided we wouldn't do that.

Yeah I agree that having a list with an empty string isn't great, but the empty string _is_ the name of the directory. We'd have to carve out an exception for OPFS roots in this (which isn't that big of a deal):

```
If [=locating an entry=] given [=/file system locator=] |locator| returns a
[=/file system entry=] |entry|, then |entry|'s [=file system entry/name=] must
be the same [=string=] as the last [=list/item=] of
|locator|'s [=file system locator/path=].
```

> For file pickers you'd want the root to be a unique identifier and the path to be a list containing the file name, I think. At least by default `<input type=file>` shouldn't start exposing more about the user's machine.

I agree that this would be ideal, but consider the following sequence:
- show a file picker and select `data/drafts/example.txt`
- show a directory picker and select `data`
- resolve `data/drafts/example.txt` relative to `data` (this should be `[ 'drafts', 'example.txt' ]`)

This won't resolve correctly if the "file system path" of the first file is just `[ 'example.txt' ]`

We could specify that the last item of the "file system path" should be the entry's name, but I think the rest of the path probably needs to be implementation-defined? And since the "file system path" isn't expected to be exposed to the end user, I don't think it should really matter?

annevk: > The only nitpick here is that if { create: true } currently fails rather than recursively create an entry. So we'd still need to check this's locator first before appending the given name, or change the behavior

Can you elaborate on this, I'm not sure I follow.

> but consider the following sequence:

Well, you could resolve with null in this case. We don't have to tell the website more than the user told them. If you want to make it work you'd have to do away with the concept of a root I think (that would have to be part of the path), but I'm not necessarily convinced it needs to work.

annevk: > Taking a step back... it seems like a bad idea to have different locators that point to the same entry anyways?

I'm not sure. I'll try to get some clarity on this.

> By the way, any thoughts on this?

It seems reasonable. We probably have to clarify somewhere that the name getter can return an invalid file name in this special case.

index.bs
```diff
@@ -674,17 +737,17 @@ The <dfn method for=FileSystemDirectoryHandle>getFileHandle(|name|, |options|)</
 
 1. Let |result| be [=a new promise=].
 1. Let |realm| be [=this=]'s [=relevant Realm=].
-1. Let |entry| be the result of [=locating an entry=]
-   given [=this=]'s [=FileSystemHandle/locator=].
+1. Let |locator| [=this=]'s [=FileSystemHandle/locator=].
```
a-sully, line 740: nit: `Let |locator| be [=this=]'s`...

here and below

index.bs
a-sully: > If we want to assert that the path above "has to be" `/data/drafts/example.txt` (in list form) then we could:
> * ensure that handles acquired from various API surfaces are given a defined path
>   * set the path of the root of an OPFS to a list containing one empty string
>   * outside of the OPFS (e.g. for a file picked from a file picker), let the path be "implementation defined"
> * create handles from existing handles by manipulating the existing handle's locator's path
>   * add new versions of the "create a new FileSystem{File|Directory}Handle" algorithms that append to the path of the existing handle

This is all included in the latest patch, aside from creating handles outside of the OPFS (which will have to be done in a follow-up on https://github.com/WICG/file-system-access, for which I _think_ I've exported all the terms I'll need... 🤞)

Also added a warning that the "file system path" shouldn't be exposed directly to the website process - it can only be implied by resolving the handle to a parent, as in this use case:

> * show a file picker and select `data/drafts/example.txt`
> * show a directory picker and select `data`
> * resolve `data/drafts/example.txt` relative to `data` (this should be `[ 'drafts', 'example.txt' ]`)

index.bs
```diff
@@ -147,35 +151,125 @@ Exactly how external changes are reflected in the data structures defined by thi
...
+a [=/file system path=] |b| if
+|a|'s [=list/size=] is the same as |b|'s [=list/size=] and
+[=list/for each=] |index| of |a|'s [=list/indices=]
+|a|.\[[|index|]] is |b|.\[[|index|]].
+
+Note: The contents of a [=/file system locator=], including its
+[=file system locator/path=], are not expected to be shared in their entirety
+with the website process. The [=/file system path=] might contain components
+which are not known to the website unless the [=/file system locator=] is later
+[=file system locator/resolved=] relative to a parent [=directory locator=].
```
annevk, line 271: I think we want `<p class=warning>` for this. I think "website process" is probably okay language for now. Although maybe we should say "agent cluster" or "agent".

index.bs
```diff
@@ -226,14 +323,17 @@ Their [=deserialization steps=], given |serialized| and |value| are:
...
 
 The <dfn attribute for=FileSystemHandle>kind</dfn> getter steps are to return
-{{FileSystemHandleKind/"file"}} if [=this=] is a [=file entry=]; otherwise
-{{FileSystemHandleKind/"directory"}}.
+[=this=]'s [=FileSystemHandle/locator=]'s [=file system locator/kind=].
 
-The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return [=this=]'s [=file system entry/name=].
+The <dfn attribute for=FileSystemHandle>name</dfn> getter steps are to return
+the empty [=string=] if [=this=] is the root directory of the
+[=origin private file system=], otherwise the last [=list/item=] (a [=string=])
```
annevk, line 335:
```suggestion
[=origin private file system=]; otherwise the last [=list/item=] (a [=string=])
```

annevk: I think "origin private file system"-ness needs to be stored on _this_ in a field somewhere or equivalent. This reads a little funky.

annevk (commented):
Thanks again for tackling this! I will attempt to give this a more detailed read through next week. Perhaps meanwhile @jesup or @asutherland can have a look?

index.bs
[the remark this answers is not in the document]

index.bs
a-sully: Hmm, yeah it's somewhat implied by the "file system root", but I agree we could make it explicit. It would be nice to give each locator an optional "storage bucket", and if a bucket is set then the object must be from an OPFS (since non-OPFS files live outside the browser, and therefore the bucket system, entirely). But storage buckets aren't yet standardized, which is why I've only mentioned them in non-normative text for now

I'm happy to add a boolean field for now, but there are some other OPFS-related changes (e.g. #92 and https://github.com/whatwg/fs/pull/9/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R1358, which I think may now be wrong) to consider that I don't want to creep into the scope of this PR... Would it be okay to leave this for a follow-up?

index.bs
a-sully: Ack. Added the warning label, and happy to update the other text if desired

index.bs
annevk: That sounds fine.

index.bs
```diff
@@ -213,7 +312,7 @@ Their [=deserialization steps=], given |serialized| and |value| are:
 1. If |serialized|.\[[Origin]] is not [=same origin=] with
    |value|'s [=relevant settings object=]'s [=environment settings object/origin=],
    then [=throw=] a "{{DataCloneError}}" {{DOMException}}.
-1. Set |value|'s [=FileSystemHandle/entry=] to |serialized|.\[[Entry]]
+1. Set |value|'s [=FileSystemHandle/locator=] to |serialized|.\[[Locator]]
```
annevk, line 315:
```suggestion
1. Set |value|'s [=FileSystemHandle/locator=] to |serialized|.\[[Locator]].
```

index.bs
```diff
@@ -289,12 +426,18 @@ A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/entry=] must be a [=f
...
+   given [=this=]'s [=FileSystemHandle/locator=].
 1. Run these steps [=in parallel=]:
-  1. Let |access| be the result of running [=this=]'s [=FileSystemHandle/entry=]'s
+  1. Let |access| be the result of running |entry|'s
      [=file system entry/query access=] given "`read`".
   1. If |access| is not "{{PermissionState/granted}}",
      [=/reject=] |result| with a "{{NotAllowedError}}" {{DOMException}} and abort.
-  1. Let |entry| be [=this=]'s [=FileSystemHandle/entry=].
+
+  1. If |entry| is `null`, [=/reject=] |result| with a
```
annevk, line 437:
```suggestion
  1. If |entry| is null, [=/reject=] |result| with a
```

index.bs
```diff
@@ -289,12 +426,18 @@ A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/entry=] must be a [=f
 The <dfn method for=FileSystemFileHandle>getFile()</dfn> method steps are:
 
 1. Let |result| be [=a new promise=].
+1. Let |entry| be the result of [=locating an entry=]
+   given [=this=]'s [=FileSystemHandle/locator=].
```
annevk, line 430: Can this happen synchronously on the main thread? This seems like something that should happen in parallel and in the future in the file system parallel queue. (This applies multiple times I think.)

(Not entirely sure how this works with async iteration. That already seems to assume it's running in parallel with how it runs "query access"...)

index.bs
```diff
@@ -271,7 +374,40 @@ interface FileSystemFileHandle : FileSystemHandle {
 };
 </xmp>
 
-A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/entry=] must be a [=file entry=].
+A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/locator=]'s
+[=file system locator/kind=] is {{FileSystemHandleKind/"file"}}.
+
+<div algorithm>
+To <dfn data-lt="creating a child FileSystemFileHandle">create a child FileSystemFileHandle</dfn>
```
annevk, line 381: Since FileSystemFileHandle is code, we should still format it as such with `<code>`. Also from the places that call this.

index.bs
```diff
@@ -271,7 +374,40 @@ interface FileSystemFileHandle : FileSystemHandle {
 };
 </xmp>
 
-A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/entry=] must be a [=file entry=].
+A {{FileSystemFileHandle}}'s associated [=FileSystemHandle/locator=]'s
+[=file system locator/kind=] is {{FileSystemHandleKind/"file"}}.
```
annevk, line 378: You could make this a note. (Unless text requires something I usually make it either a note or an example. If you think this is a better you should feel free to leave it as-is though. It's definitely an improvement over what was here before.)

index.bs
```diff
@@ -1303,8 +1540,18 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
   1. Set |dir|'s [=directory entry/children=] to an empty [=/set=].
   1. Set |map|["root"] to |dir|.
 
-1. Return [=a promise resolved with=] a new {{FileSystemDirectoryHandle}},
-   whose associated [=FileSystemHandle/entry=] is |map|["root"].
+1. Let |root| be an [=implementation-defined=] opaque [=string=].
+1. Let |path| be an empty [=/list=].
```
annevk, line 1544:
```suggestion
1. Let |path| be « ».
```

index.bs
```diff
@@ -147,35 +151,127 @@ Exactly how external changes are reflected in the data structures defined by thi
...
 
+A <dfn export>file system locator</dfn> represents a potential location of a
+[=/file system entry=]. A [=/file system locator=] is either a [=file locator=]
+or a [=directory locator=].
+
+<!-- TODO(a-sully): Consider giving each locator a Storage Bucket. -->
+
+Each [=/file system locator=] has an associated <dfn export for="file system locator" id=locator-path>path</dfn> (a [=/file system path=]),
+a <dfn export for="file system locator" id=locator-kind>kind</dfn> (a {{FileSystemHandleKind}}), and
+a <dfn export for="file system locator" id=locator-root>root</dfn> (an opaque [=string=]).
```
annevk, line 201:
```suggestion
a <dfn export for="file system locator" id=locator-root>root</dfn> (a [=file system root=]).
```

annevk (commented):
Overall this looks pretty good to me. I found a couple of remaining nits. Can you please file the follow-ups so we can point to them?

index.bs
a-sully: Seems reasonable to make this a note

index.bs
a-sully: Hmm looks like we do the same thing for FileSystemWritableFileStream and FileSystemSyncAccessHandle. I'll fix those in a follow-up

index.bs
a-sully: Good point. I've moved this into the "in parallel" section, which we can then deal with in #95

annevk:
Looking at this again I'm not sure I understand the setup around name.

* Do entries need a name at all?
* Why does `getDirectory()` need to set a name? Can it not just set path to "« the empty string »"? That would also allow us to simplify the getter steps as they would not have to special case OPFS anymore.

I pushed a cleanup commit that removes some of the wording around name that seemed to duplicate other requirements. And where it was used incorrectly as a type. I think we can go further, but I'm not entirely sure how to reword the various children iterators.

a-sully:
> I pushed a cleanup commit that removes some of the wording around name that seemed to duplicate other requirements. And where it was used incorrectly as a type

This looks good, assuming we agree below on how to handle the OPFS root. Thanks!

> Looking at this again I'm not sure I understand the setup around name.
> 
> * Do entries need a name at all?
>
>  I'm not entirely sure how to reword the various children iterators.

Are we sure we need to reword the various children iterators? It seems fine to use entries (as opposed to locators) for file system operations like reading the contents of an entry (i.e. the bytes of a file, or the children of a directory).

We need some way to [identify the children within a directory](https://fs.spec.whatwg.org/#directory). Iterating through a directory doesn't make much sense if the files within it don't have a name...

If the intention is that an "entry" represents an OS-level file/directory, then giving each entry a name (and contents, as mentioned above) is probably what we want. Alternatively, we could consider giving each entry a path... but that's not really how most file systems work. You can later get a file's path by walking up the directory chain, but the path isn't inherent to the file itself.

> * Why does `getDirectory()` need to set a name? Can it not just set path to "« the empty string »"? That would also allow us to simplify the getter steps as they would not have to special case OPFS anymore.

Yeah setting the path to "« the empty string »" is what I'd proposed earlier, so I'm on board with that :) https://github.com/whatwg/fs/pull/96#discussion_r1125034492

The entry itself still needs a name though, as per above. And we need `|map|["root"]` to point to a directory entry that exists, not a locator that may not. Right?

annevk:
Okay, that sounds reasonable. I guess that means we should keep the path[last]/name invariant in some form, probably as new bullet points in the lists I created. If you'll indulge me, what happens when you hold an entry named `x` in JS and someone does `mv x y` on the command line?

index.bs
```diff
@@ -147,35 +151,135 @@ Exactly how external changes are reflected in the data structures defined by thi
...
+|a|'s [=file system locator/kind=] is |b|'s [=file system locator/kind=],
+|a|'s [=file system locator/root=] is |b|'s [=file system locator/root=], and
+|a|'s [=file system locator/path=] is [=the same path as=] |b|'s [=file system locator/path=].
+
+<div algorithm>
+The <dfn export data-lt="locating an entry">locate an entry</dfn> algorithm given a
+[=/file system locator=] |locator| runs an [=implementation-defined=] series of steps adhering to
+these constraints:
+
+<ul>
```
a-sully, line 230: Thoughts on using standard markdown formatting for 
```
- lists and
  - sub-lists
```
?

This spec mostly avoids HTML if markdown suffices. Ditto with

```
create a new <code>FileSystemFileHandle</code>
create a new `FileSystemFileHandle`
```

annevk: I think that's fine. That's really your call as editor. I wasn't sure about the syntax so I resorted to something I'm comfortable with.

a-sully: SGTM, just wanted to confirm before messing with your changes :)

PR #108, opened 2023-04-03
Add explainers/ folder

a-sully:
It's an open question as to where explainers related to this specification should live. With a handful of explainers in existence and a number more in the works, it would be nice to establish a pattern for where these documents should live.

The current pattern of creating un-merged PRs not sustainable. See https://github.com/w3ctag/design-reviews/issues/805#issuecomment-1467640227

The most common options seem to be (see some previous discussion https://github.com/whatwg/fs/pull/46#issuecomment-1230531707):
- _Posting explainers to personal forks (this seems to be the most common pattern)_
  - Pros:
    - The upstream repo remains uncluttered
  - Cons:
    - This can lead to broken links (e.g. https://github.com/whatwg/storage/pull/139)
    - While the spec provides the _how_ and the _what_, it can be quite challenging to trace the paper trail of _why_ the spec is the way it is, especially if explainers are coming from a variety of contributors
- _Posting explainers into the specification directly into the spec repo_
  - Pros:
    - Easily find all the explainers related to a specification
  - Cons:
    - The root of the spec can become quite cluttered if you have a lot of explainers
    - It can lead to the anti-pattern of putting the useful text _only_ in the explainers and not in the spec text itself

This PR proposes a new option: _Posting explainers into an `explainers/` folder in the spec repo_

As a fairly new specification, there are a known, small(ish) number of explainers such that it's not hard to aggregate them for now (with some more in https://github.com/WICG/file-system-access that we may want to consider porting parts of over eventually). But this spec has lots of room to grow, and I expect a number of explainers coming in the near future. I personally think it would be nice (and I suspect others, such as those providing TAG reviews, might agree) to have all these explainers in an easy-to-find location, without cluttering the spec repo too much

This PR merges the following explainers:
- https://github.com/a-sully/fs/pull/1
- https://github.com/a-sully/fs/pull/2

https://github.com/whatwg/fs/pull/46 is not merged for now, as to not lose the discussions on that PR already

cc @torgo


a-sully:
@annevk @domenic does this approach seem reasonable?

domenic:
Yes, this seems reasonable to me. As I noted previously, I think explainer organization should be up to the editors, e.g. in whatwg/streams we've put them in the root, whereas for many of @annevk's specs he's preferred leaving them in other repos.

annevk:
I think the problem I have with them is that they are quite official-looking and people might mistake them for WHATWG-endorsed ideas.

a-sully:
Fair point, though personally I think I'd still prefer to have all these proposals in one place. There are some nice benefits to this organization (e.g. the timeline of feature additions, and _why_ features were developed, is much easier to trace and therefore much more transparent) and I think we can mitigate the risks you mentioned by presenting these in a less official-sounding way.

In that vein, I've renamed this folder to `proposals/`, which I think better conveys what these documents are anyways

annevk:
I think that still doesn't really convey their insignificance to the WHATWG process. It seems better to keep them out-of-band.

domenic:
Perhaps a README.md in the subdirectory could help assuage @annevk's concerns.

Issue #123, opened 2023-05-23
File and Directory Watch Functionality for the FileSystem API


WEBPerformace:
Hello,

I'd like to suggest the addition of a feature to the FileSystem API that would allow watching for changes in files and directories. 

I've seen it mentioned many times in the comments for issues in this repository that are not directly related to this problem, but I did not find a specific issue specifically for that.

This ability is an essential component in many types of applications, including IDEs, file managers, photo/video editors, ML dataset managers, syncing apps, and any web app that uses its own type of file browser.

The lack of a file system API that includes watching capabilities can limit the effectiveness of native web apps, pushing developers towards heavier client applications or Electron-based solutions. Many projects could become PWAs instead of heavier client applications if this feature was implemented.

There are workaround solutions currently being used, such as maintaining a local list of files and checking that against the directory listing or recursively iterating over all files. However, these solutions can be resource-intensive and aren't as efficient or reliable as native events could be.

The need for this feature extends beyond web apps. For example, a i've seen expressed interest in being notified when a device like a USB hard disk or pen is plugged into the system. Additionally, this would significantly benefit browser-based debuggers.

Most OSes include their APIs for watching file changes, such as `ReadDirectoryChangesW` on Windows or the File System Events API on MacOS. Leveraging these APIs could potentially reduce the time needed for implementation.

Thank you for considering this feature request. The addition of a file and directory watch functionality to the FileSystem API would undoubtedly empower developers to create more robust and efficient web apps.

a-sully:
I agree this is useful! This idea has extensive prior discussion at https://github.com/WICG/file-system-access/issues/72, though that discussion is focused primarily on watching the local file system, while I believe this capability is also useful for the Bucket File System (a.k.a. Origin Private File System)

A proposal for a new `FileSystemObserver` interface is here: https://github.com/whatwg/fs/pull/124 (I expect to merge the PR soon)

PR #124, opened 2023-06-06
Add proposal for the FileSystemObserver interface

a-sully:
Proposal for a new interface which allows a website to be notified of changes to files and directories.

See #123 and https://github.com/WICG/file-system-access/issues/72

### DIFF
--- a/proposals/FileSystemObserver.md
+++ b/proposals/FileSystemObserver.md
@@ -1,0 +1,681 @@
+# The FileSystemObserver Interface
+
+## Authors
+
+*   [Austin Sullivan](asully@chromium.org) (Google)
+
+## Participate
+
+*   [Issue tracker](https://github.com/whatwg/fs/issues)
+
+## Introduction
+
+The file system is a shared resource that can be modified from several contexts. A [Bucket File System](https://fs.spec.whatwg.org/#sandboxed-filesystem) spans numerous [agents](https://tc39.es/ecma262/#sec-agents) - tabs, workers, etc - within the same [storage key](https://storage.spec.whatwg.org/#storage-keys). The local file system also spans across origins and other applications on the host operating system.
+
+For a given agent to know about modifications to the file system - made either by itself or from some external context - it can currently poll the file system to detect changes. This is inefficient and does not scale well.
+
+This explainer proposes a `FileSystemObserver` interface which will much more easily allow a website to be notified of changes to the file system.
+
+## Goals
+
+*   Simplify application logic and improve the ergonomics of watching file paths
+*   Improve the efficiency of watching file paths on the local file system
+*   Provide best-effort information of changes to the local file system
+*   Guarantee consistent behavior across platforms with regards to the contents of a file system change record for a corresponding change to a Bucket File System
+
+## Non-Goals
+
+*   Expose any information to the web that isn’t already exposed
+*   Expand the permissions of a website as a result of a file system change
+*   Provide notification of changes that occur outside the scope of a `FileSystemObserver` connection (e.g. before the `FileSystemObserver` is created or after the tab is closed)
+*   Guarantee that all file system changes which occur while a `FileSystemObserver` is connected are reported. See [Guaranteeing that Changes are Not Missed](#guaranteeing-that-changes-are-not-missed)
+*   Guarantee consistent behavior across platforms with regards to the contents of a file system change record for a corresponding change to the local file system. See [Cross Platform Compatibility](#cross-platform-compatibility)
+
+## Use Cases
+
+*   [Binding a UI element](https://web.dev/indexeddb-uidatabinding/) to the contents of a file
+*   Notifying the main thread of file system changes from a worker or another tab
+*   Syncing file system changes to a server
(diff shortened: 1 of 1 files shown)


jespertheend:
Thanks for this proposal!
My application is currently not polling (for performance reasons) but traversing the file tree when the window receives focus.
A big use case for me is to make external changes to shader code and immediately see the results as you type. Which is currently not possible because you have to focus the browser window after every change.
So this proposal is certainly much appreciated!

I'm not a code owner so I might not be the right person for this. But these are some thoughts I had after reading the proposal.

- When some browsers point to the same `FileSystemHandle` reference as one that already exists. Some developers might compare two handles using the equality operator rather than using `isSameEntry()`. They might then test their application in one browser, while another might not behave this way. To improve compatibility I think it would be good to explicitly create new `FileSystemHandle` instances for every `FileSystemChangeRecord`.
- You mentioned that linux has no native support for recursive watches. Would this mean that changes to files in subdirectories are not reported at all? Or would it only mean that the `FileSystemChangeType` might not be as accurate?
- Say a user agent supports the `FileSystemObserver` on some directories, but not on a specific directory.
An application developer would need a way to know whether to fall back on polling or not. The `"errored"` `FileSystemChangeType` could work for this, but I'm worried that without explicitly mentioning what error should be thrown on the `observe()` call, that some browsers might end up with situations where simply no events will be fired at all.

a-sully:
> Thanks for this proposal! My application is currently not polling (for performance reasons) but traversing the file tree when the window receives focus. A big use case for me is to make external changes to shader code and immediately see the results as you type. Which is currently not possible because you have to focus the browser window after every change. So this proposal is certainly much appreciated!

Thanks for taking a look at the proposal! I expect it will address your use case :)

> I'm not a code owner so I might not be the right person for this. But these are some thoughts I had after reading the proposal.
> 
> * When some browsers point to the same `FileSystemHandle` reference as one that already exists. Some developers might compare two handles using the equality operator rather than using `isSameEntry()`. They might then test their application in one browser, while another might not behave this way. To improve compatibility I think it would be good to explicitly create new `FileSystemHandle` instances for every `FileSystemChangeRecord`.

Could you file a separate issue for this with a repro case? I did some basic tests and wasn't able to get `==` or `===` to return true for two handles pointing to the same file (i.e. where `isSameEntry()` returns true).

We should probably add web platform tests, especially if there is a discrepancy https://github.com/web-platform-tests/wpt/tree/master/fs

> * You mentioned that linux has no native support for recursive watches. Would this mean that changes to files in subdirectories are not reported at all? Or would it only mean that the `FileSystemChangeType` might not be as accurate?

The user agent will handle watching subdirectories. However, it is _significantly_ more resource-intensive (i.e. consumes memory, and more importantly file descriptors) to recursively watch a directory using [inotify](https://man7.org/linux/man-pages/man7/inotify.7.html), since a watch descriptor needs to be created for every subdirectory. This is an implementation detail that web authors ideally wouldn't need to worry about, but given the exponential difference in system resource use and that user agents will add restrictions on the number of (sub)directories that can be watched (and behavioral differences, such as that [changes to a newly-mounted sub-directory could be missed while the watch descriptors are being created](https://man7.org/linux/man-pages/man7/inotify.7.html#:~:text=Inotify%20monitoring%20of,that%20it%20contains.)) it's unfortunately a detail that's hard to hide... 

For example, on Linux it's much more likely that `observer.observe(dirHandleWithManySubDirs, { recursive: true })` will fail (by hitting some limit set by either the user agent or OS) than `observer.observe(dirHandleWithManySubDirs, { recursive: false })`

> * Say a user agent supports the `FileSystemObserver` on some directories, but not on a specific directory.
>   An application developer would need a way to know whether to fall back on polling or not. The `"errored"` `FileSystemChangeType` could work for this, but I'm worried that without explicitly mentioning what error should be thrown on the `observe()` call, that some browsers might end up with situations where simply no events will be fired at all.

Unfortunately, it's hard for the user agent to know much about why a watch has failed, either. Your best bet is to attempt to restart the observation by calling `observe()` again, which, if it fails, may throw a useful exception (e.g. a `NotAllowedError` if permissions were lost and you need to call `requestPermission()`)

inexorabletash (approved):

annevk:
As a reminder, changes like this shouldn't be committed without a `Meta: ` prefix as per the commit message guidelines. It's not worth it to rewrite main for this, but please let's not do it again.

jespertheend:
> I did some basic tests and wasn't able to get == or === to return true for two handles

Ah no this was just a hypothetical situation in which a user agent might reuse existing references in `FileSystemChangeRecord`s. But you're right, web platform tests would probably take care of that.

> The user agent will handle watching subdirectories. However, it is significantly more resource-intensive (i.e. consumes memory, and more importantly file descriptors) to recursively watch a directory using [inotify](https://man7.org/linux/man-pages/man7/inotify.7.html), since a watch descriptor needs to be created for every subdirectory. This is an implementation detail that web authors ideally wouldn't need to worry about, but given the exponential difference in system resource use and that user agents will add restrictions on the number of (sub)directories that can be watched (and behavioral differences, such as that [changes to a newly-mounted sub-directory could be missed while the watch descriptors are being created](https://man7.org/linux/man-pages/man7/inotify.7.html#:~:text=Inotify%20monitoring%20of,that%20it%20contains.)) it's unfortunately a detail that's hard to hide...

Thanks for the clarification, that makes sense.

PR #129, opened 2023-06-14
Rename the Origin Private File System to the Bucket File System

### LINKED ISSUE #92: Consider renaming " the Origin Private File System" in anticipation of Storage Buckets

a-sully:
Fixes #92

***
<a href="https://whatpr.org/fs/129.html" title="Last updated on Jun 16, 2023, 9:40 PM UTC (4fadcaf)">Preview</a> | <a href="https://whatpr.org/fs/129/d636c34...4fadcaf.html" title="Last updated on Jun 16, 2023, 9:40 PM UTC (4fadcaf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -254,7 +254,7 @@ A <dfn export>file system root</dfn> is an opaque [=string=] whose value is
 <p class=example id=example-locator>For a [=/file system locator=] |locator|
 whichs [=locate an entry|locates to=] a [=file entry=] |entry| that conceptually
 exists at the path `data/drafts/example.txt` relative to the root directory of
-the [=origin private file system=],
+a [=/bucket file system=],
 |locator|'s [=file system locator/kind=] has to be {{FileSystemHandleKind/"file"}},
 |locator|'s [=file system locator/path=] has to be « "`data`", "`drafts`", "`example.txt`" », and
 |locator|'s [=file system locator/root=] might include relevant identifying
@@ -342,12 +342,12 @@ Note: Multiple {{FileSystemHandle}} objects can have
 [=the same locator as|the same=] [=/file system locator=].
 
 A {{FileSystemHandle}}
-<dfn for=FileSystemHandle export>is in an origin private file system</dfn>
+<dfn for=FileSystemHandle export>is in a bucket file system</dfn>
 if the first [=list/item=] of its [=FileSystemHandle/locator=]'s
 [=file system locator/path=] is the empty string.
 
-Note: This is a bit magical, but it works since only the root directory of an
-[=origin private file system=] can have a [=file system locator/path=] which
+Note: This is a bit magical, but it works since only the root directory of a
+[=/bucket file system=] can have a [=file system locator/path=] which
 [=list/contains=] an empty string. See {{StorageManager/getDirectory()}}.
 All other [=list/item=]s of a [=file system locator/path=] will be a
 [=valid file name=].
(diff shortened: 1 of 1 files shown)


a-sully:
cc @annevk @asutherland @inexorabletash @mkruisselbrink

tomayac:
Since the publicized name in blogs etc. has been [origin private file system](https://www.google.com/search?q=%22origin+private+file+system%22), I wonder what we do "brand"-wise. Would it be fair to keep saying "origin private file system" in general, and then more concretely say that "the  _origin private file system_ can contain one or multiple _bucket file systems_"?

a-sully:
I think it's fine to continue to refer to it publicly as the  _Origin Private File System_, which at this point seems to have some name recognition and has even crept into interfaces such as [sqlite3 WASM](https://sqlite.org/wasm/doc/trunk/persistence.md#vfs-opfs). But from the perspective of the spec, we probably should use the more specific term? Developers presumably look at MDN rather than the spec itself anyways, and MDN could have a note explaining the discrepancy

Alternatively we could just leave things as they are (or perhaps make the minimal change of saying ["an OPFS" instead of "the OPFS"](https://github.com/whatwg/fs/issues/92#issue-1533031756), to imply that there could be multiple per origin) 

> "the origin private file system can contain one or multiple bucket file systems"

I wouldn't suggest framing it this way, since it makes it seem like an OPFS contains several file systems. An _origin_ (or more accurately, a storage key) can contain several buckets, and therefore several file systems (one per bucket)

index.bs
```diff
@@ -1669,20 +1672,20 @@ guarantee.
 </div>
 
 
-# Accessing the Origin Private File System # {#sandboxed-filesystem}
+# Accessing the Bucket File System # {#sandboxed-filesystem}
 
-The <dfn export>origin private file system</dfn> is a [=storage endpoint=] whose
+The <dfn export>bucket file system</dfn> is a [=storage endpoint=] whose
```
annevk, line 1677: We should keep the existing ID. (For the other `<dfn>` change I don't think it matters since we introduced it two days ago, but this is much more likely to be referenced already.)

And FWIW, in my opinion it's fine if the ID has a different name from what it identifies. It nicely shows the specification has evolved.

annevk (approved):
Looks good modulo nit.

annevk:
No, Bucket here refers to Storage Standard terminology. Not the Storage Buckets APIs.

tomayac:
> No, Bucket here refers to Storage Standard terminology. Not the Storage Buckets APIs.

Thank you! I see, as per this [definition](https://storage.spec.whatwg.org/#storage-bucket), which the Storage Buckets API also [references](https://wicg.github.io/storage-buckets/#storage-bucket-getdirectory:~:text=Let%20bucket%20be%20a%20new%20storage%20bucket%20with%20name%20name.). It's just a bit confusing that there's also a [`StorageBucket`](https://wicg.github.io/storage-buckets/#storagebucket).

annevk:
Well, that's an API instance of that concept, so it's quite logical in a way.

index.bs
a-sully: Done. I was ready to update all the refs anyways (see https://github.com/WICG/file-system-access/pull/416) but it is nice to see the evolution :)

PR #131, opened 2023-06-15
Add a serializable "permission root locator" to a FileSystemHandle

### LINKED ISSUE #101: Make access check algorithms no longer associated with an entry
Access checks are currently tied to a "file system entry". As of #96, a "file system entry" now corresponds to an actual file or directory on disk. We need to support access checks on FileSystemHandles which don't correspond to an entry (e.g. the entry has been removed), so the access checks must be tied to something else.

We could tie access checks to a FileSystemHandle or a "file system locator". See https://github.com/whatwg/fs/pull/96#discussion_r1116465722

a-sully:
Until now, access checks were tied to a "file system entry". As of #96, a "file system entry" corresponds to an actual file or directory on disk. We need to support access checks on FileSystemHandles which don't correspond to an entry (e.g. the entry has been removed), so the access checks must be tied to something else.

This PR gives each `FileSystemHandle` a "permission root locator". A newly created child `FileSystemHandle` will inherit this attribute from its parent (as opposed to its access algorithms), such that if `subDir` was created from `rootDir`, `subDir.requestPermission()` is equivalent to `rootDir.requestPermission()`. This "permission root locator" is serialized, such that storing a handle in IDB now retains information about where the handle originated from

The new "query access" and one "request access" algorithms are expected to be overridden in https://wicg.github.io/file-system-access/#permissions to handle non-BucketFileSystem use cases.

Fixes #101

***
<a href="https://whatpr.org/fs/131.html" title="Last updated on Jul 18, 2023, 4:09 AM UTC (b29ed88)">Preview</a> | <a href="https://whatpr.org/fs/131/bee4f50...b29ed88.html" title="Last updated on Jul 18, 2023, 4:09 AM UTC (b29ed88)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -285,13 +285,11 @@ which are not known to the website unless the [=/file system locator=] is later
 [=file system locator/resolved=] relative to a parent [=directory locator=].
 
 <div algorithm>
-The <dfn id=entry-query-access lt="querying access" export>file system query access algorithm</dfn>
-is used for querying access to a location in the file system. It takes
-a [=/file system locator=] |locator| and
-a "`read`" or "`readwrite`" <var ignore>mode</var>
-and returns a [=/file system access result=].
-Unless specified otherwise, it runs these steps:
+<dfn abstract-op id=entry-query-access lt=QueryFileSystemAccess>QueryFileSystemAccess(|locator|, |mode|)</dfn>
+runs these steps, unless specified otherwise:
 
+1. Assert: |locator| is a [=/file system locator=].
+1. Assert: |mode| is "`read`" or "`readwrite`".
 1. If |locator| [=file system locator/is in a bucket file system=]
    return a [=/file system access result=] with
    a [=file system access result/permission state=]
@@ -306,17 +304,15 @@ Note: These steps have to be run on the [=file system queue=].
 
 </div>
 
-The <dfn id=entry-request-access lt="requesting access" export>file system request access algorithm</dfn>
-is the same as the [=/file system query access algorithm=],
+<dfn abstract-op id=entry-request-access lt=RequestFileSystemAccess>RequestFileSystemAccess(|locator|, |mode|)</dfn>
+returns [$QueryFileSystemAccess$](|locator|, |mode|),
 unless specified otherwise.
 
-Note: Dependent specifications may override the
-[=/file system query access algorithm=] and
-[=/file system request access algorithm=].
+Note: Dependent specifications may override
+[$QueryFileSystemAccess$] and [$RequestFileSystemAccess$].
 
 A <dfn export>file system access result</dfn> is a [=struct=] encapsulating the
-result of a [=/file system query access algorithm=] or a
-[=/file system request access algorithm=].
+result of running [$QueryFileSystemAccess$] or [$RequestFileSystemAccess$].
 It has the following [=struct/items=]:
 
 : <dfn for="file system access result">permission state</dfn>
@@ -332,13 +328,17 @@ It has the following [=struct/items=]:
 
 <p class=warning> Dependent specifications may consider this API a
 [=powerful feature=]. However, unlike other [=powerful features=] whose
-[=permission request algorithm=] may throw, the
-[=/file system query access algorithm=] and [=/file system request access algorithm=]
+[=permission request algorithm=] may throw,
+[$QueryFileSystemAccess$] and [$RequestFileSystemAccess$]
 each must run [=in parallel=] on the [=file system queue=] and are
 therefore not allowed to throw. Instead, the caller is expected to
 [=queue a storage task=] to [=/reject=], as appropriate,
 should these algorithms return an [=file system access result/error name=]
 other than the empty string.
+
+<p class=example id=example-query-access>To check whether a website can access a
+{{FileSystemHandle}} |handle| with |mode|, run
+[$QueryFileSystemAccess$](|handle|'s [=FileSystemHandle/permission root locator=], |mode|).
 
 ## The {{FileSystemHandle}} interface ## {#api-filesystemhandle}
 
@@ -474,9 +474,9 @@ in a [=/Realm=] |realm|:
    |parentPermissionRootLocator|.
 
    Note: This ensures that permissions are inherited from parent to child.
-   [=Querying access=] or [=requesting access=] to |handle| will
-   [=query access=] or [=request access=] to its parent - or its parent's
-   parent, and so forth - as appropriate.
+   Querying access or requesting access to |handle| will
+   query access or request access to its parent - or its parent's parent,
+   and so forth - as appropriate.
 
 1. Return |handle|.
 
@@ -521,8 +521,8 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method steps are:
    [=this=]'s [=FileSystemHandle/permission root locator=].
 1. Let |global| be [=this=]'s [=relevant global object=].
 1. [=Enqueue the following steps=] to the [=file system queue=]:
-  1. Let |accessResult| be the result of [=querying access=] given
-     |permissionRootLocator| and "`read`".
+  1. Let |accessResult| be the result of running
+     [$QueryFileSystemAccess$](|permissionRootLocator|, "`read`").
   1. Let |entry| be the result of [=locating an entry=] given |locator|.
 
   1. [=Queue a storage task=] with |global| to run these steps:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -488,9 +486,9 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method steps are:
 1. Let |locator| be [=this=]'s [=FileSystemHandle/locator=].
 1. Let |global| be [=this=]'s [=relevant global object=].
 1. [=Enqueue the following steps=] to the [=file system queue=]:
+  1. Let |accessResult| be the result of running |global|'s
```
a-sully, line 489: Q: is use of `global` correct here? (and throughout this PR)

index.bs
```diff
@@ -637,6 +634,7 @@ The <dfn method for=FileSystemFileHandle>createSyncAccessHandle()</dfn> method s
      [=/reject=] |result| with an "{{InvalidStateError}}" {{DOMException}} and
      abort these steps.
 
+  1. Let |entry| be the result of [=locating an entry=] given |locator|.
```
a-sully, line 637: FYI: I've re-ordered the "locating an entry" statements in several algorithms, since the |entry| is no longer needed for the access checks

index.bs
```diff
@@ -1147,19 +1147,24 @@ Similarly, when piping a {{ReadableStream}} into a {{FileSystemWritableFileStrea
...
+   and returns the result of running the [=write a chunk=] algorithm with
+   |fileHandle|, |stream|, and |chunk|.
 1. Let |closeAlgorithm| be these steps:
    1. Let |closeResult| be [=a new promise=].
    1. [=Enqueue the following steps=] to the [=file system queue=]:
-      1. Let |accessResult| be the result of running |file|'s
-         [=file system entry/query access=] given "`readwrite`".
+      1. [=Assert=]: [=locating an entry=] given |fileHandle|'s
+         [=FileSystemHandle/locator=] returns a [=file entry=] that is
+         [=the same entry as=] |fileEntry|.
```
a-sully, line 1163: We should be guaranteed of this since the file entry is locked and modifying operations such as `move()` and `remove()` (will) take a lock (once they're specified). Do you think this is worth `Note: `ing?

a-sully:
cc @mkruisselbrink but he's OOO for a bit so sending over to @annevk for review. Thanks!

index.bs
```diff
@@ -355,6 +307,52 @@ All other [=list/item=]s of a [=file system locator/path=] will be a
 Issue(109): Consider improving this situation by giving each locator a
 [=storage bucket=].
 
+Each {{FileSystemHandle}} has an associated
+<dfn for=FileSystemHandle export>query access</dfn>
```
annevk, line 311: Please preserve IDs.

index.bs
annevk: No, the algorithm sits on a FileSystemHandle, right? So I'm not sure how this can work. It's already a bit odd to access a FileSystemHandle in parallel. Maybe it would be better if they were associated with the locator? Or does that end up doing the wrong thing?

index.bs
annevk: An Assert is a good way to note things. :-)

index.bs
a-sully: The only issue with using the locator is that it doesn't contain the context from which the handle was acquired. For example, consider this scenario:

1. The user selects `~/Documents/foo/bar.txt` from the file picker as `file`
2. The user selects `~/Documents/foo/` from the directory picker, then the site calls `dir.getFileHandle('bar.txt')` to get `fileFromDir`

Note that in this case, `file.isSameEntry(fileFromDir)` is true and the respective file system locators are the same

However, if `file.requestPermission()` is called, the "request access" algorithm corresponds to  `~/Documents/foo/bar.txt`

Meanwhile, if `fileFromDir.requestPermission()` is called, the "request access" algorithm (at least in Chromium browsers) corresponds to `~/Documents/foo/`. This is because the [`getFileHandle()` method](https://fs.spec.whatwg.org/#dom-filesystemdirectoryhandle-getfilehandle) passes down its access check algorithms to its children (see step 6). When serializing a `FileSystemHandle`, our implementation actually [includes the paths of both the parent (which was selected from the picker) and the child](https://crsrc.org/c/content/browser/file_system_access/file_system_access.proto) in the serialization.

Given all this, I see three paths forward:

- Argue that this is all just an implementation detail (e.g. it's up to the user agent whether `fileFromDir.requestPermission()` should request permission to  `~/Documents/foo/` or `~/Documents/foo/bar.txt`), and for the purposes of the spec we should just tie access checks to a "fie system locator" and call it a day
- Give each `FileSystemHandle` two locators - one for itself and one for the root it was created from. Access checks for any handle use the root's locator. Serializing a handle requires serializing both of these locators
- Give each `FileSystemHandle` some attribute that encapsulates its access check algorithms (e.g. "file system permission state"). This is passed down to all created children similar to what is currently specified for the access check algorithms (see [step 6 of `getFileHandle()`](https://fs.spec.whatwg.org/#dom-filesystemdirectoryhandle-getfilehandle)). Access checks are performed based only on this attribute, and the attribute would need to be serialized (though I'm not sure exactly how that would look)

Thoughts?

a-sully: Looking at the WICG spec more closely... (3) could look like giving a `FileSystemHandle` a `FileSystemPermissionDescriptor` member (though I'm not sure if it's a problem that a descriptor is currently specified to have a handle)

I guess there's still:

4. Tie access check algorithms to the `FileSystemHandle` itself, if there's a way to access a `FileSystemHandle` in parallel (as the WICG spec [currently does by incorrectly using `this` directly](https://wicg.github.io/file-system-access/#dom-filesystemhandle-querypermission)). As @mkruisselbrink pointed out, though, it's unclear how this permission state could be serialized

FWIW this "inherit permissions from the parent" behavior is also specified in step 3 of the [permission state constraints](https://w3c.github.io/permissions/#dfn-permission-state-constraints) algorithm for a [`FileSystemPermissionDescriptor`](https://wicg.github.io/file-system-access/#dictdef-filesystempermissiondescriptor). That text's use of "parent" was made outdated by #96, though (i.e. it can no longer be null). It seems that we should either introduce the concept of a "permission root" (as options 2 and 3 do explicitly and 4 does implicitly) or say that the context from which the handle was acquired is not relevant and the locator is all that matters (i.e. that `file.requestPermission()` should always request permission to `~/Documents/foo/`)

On another note... given that that constraint exists, I wonder if updating serialization steps might be unnecessary, since it's just a mechanism used to enforce this constraint?

Lots of options here. Eager to hear thoughts from others :)

annevk: Given that what constraint exists?

I think you're correct that locator alone doesn't work. Presumably if you do `subDir.requestPermission()` it should still impact `rootDir`? I think that argues for either storing a reference to the root handle or the root locator that a potential access algorithm could then use.

index.bs
a-sully: no longer relevant as of the latest patch

index.bs

index.bs
a-sully: In the latest patch I've given each `FileSystemHandle` a "permission root locator" (let me know if you have a better name for this). As mentioned above, this matches how Chrome serializes a handle today.

In this model, creating a handle through one of the API's entry points (which in this spec is only `navigator.storage.getDirectory()`, but also includes e.g. the picker methods in https://wicg.github.io/file-system-access/) will set the "permission root locator" to the "locator" of the handle itself. Meanwhile, any children which are created from this handle will be passed the permission root, such that `subDir.requestPermission()` is the same as `rootDir.requestPermission()`

A key advantage of this approach (as opposed to passing the access algorithm itself from parent to child) is that it's serializable. The old approach tied "query access" and "request access" algorithms to an entry (and an version of this PR tied these access algorithms to the `FileSystemHandle` itself). But these algorithms are (presumably? Please let me know if that's incorrect) not serializable, so `postMessage()` or storing a handle in IDB would lose its "where did I originate from" information

As such, there is now just one "query access" and one "request access" algorithm - which return "granted" for locators in a bucket file system; otherwise "denied" - that I intend to override in https://wicg.github.io/file-system-access/#permissions to handle the non-BucketFileSystem use cases. Does that seem like a reasonable approach?

index.bs
```diff
@@ -318,6 +284,62 @@ with the website process. The [=/file system path=] might contain components
 which are not known to the website unless the [=/file system locator=] is later
 [=file system locator/resolved=] relative to a parent [=directory locator=].
 
+<div algorithm>
+The <dfn id=entry-query-access lt="querying access" export>file system query access algorithm</dfn>
```
a-sully, line 288: Q: would this make more sense as an abstract algorithm?

index.bs
```diff
@@ -318,6 +284,62 @@ with the website process. The [=/file system path=] might contain components
 which are not known to the website unless the [=/file system locator=] is later
 [=file system locator/resolved=] relative to a parent [=directory locator=].
 
+<div algorithm>
+<dfn abstract-op id=entry-query-access lt=QueryFileSystemAccess>QueryFileSystemAccess(|locator|, |mode|)</dfn>
```
a-sully, line 288: I don't have much experience with abstract algorithms, so please let me know whether it makes sense to use in this case

annevk: I think `abstract-op` is primarily for JavaScript and some legacy cases. For new cases we can just define an algorithm as we already do elsewhere.

> To query file system access, given ...

annevk (commented):
Might be good to get feedback from someone else on the substance of this PR. Just gave some feedback on whether to use an abstract operation.

PR #134, opened 2023-06-20
Editorial: Properly format enum strings

### LINKED ISSUE #133: Editorial: string formatting
I'm pretty sure

> {{WriteCommandType/"write"}}

needs to be

> "{{WriteCommandType/write}}"

a-sully:
Before: `{{WriteCommandType/"write"}}`
After: `"{{WriteCommandType/write}}"`

Fixes #133

***
<a href="https://whatpr.org/fs/134.html" title="Last updated on Jun 20, 2023, 9:42 PM UTC (cee9c6e)">Preview</a> | <a href="https://whatpr.org/fs/134/69c51d3...cee9c6e.html" title="Last updated on Jun 20, 2023, 9:42 PM UTC (cee9c6e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -244,9 +244,9 @@ a <dfn export for="file system locator" id=locator-root>root</dfn> (a [=file sys
 Issue(109): Consider giving each locator a [=storage bucket=].
 
 A <dfn export>file locator</dfn> is a [=/file system locator=] whose
-[=file system locator/kind=] is {{FileSystemHandleKind/"file"}}.
+[=file system locator/kind=] is "{{FileSystemHandleKind/file}}".
 A <dfn export>directory locator</dfn> is a [=/file system locator=] whose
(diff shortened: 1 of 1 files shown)


a-sully:
Question: should strings like `"readwrite"` be converted to `"{{FileSystemPermissionMode/readwrite}}"`? Presumably that would require moving the declaration of [`FileSystemPermissionMode`](https://wicg.github.io/file-system-access/#enumdef-filesystempermissionmode) to this spec, though. Not sure if that's something we want to do

inexorabletash:
I defer to Anne of course, but if they're internal state not enum members, then I'd keep them as strings. I'd use https://fetch.spec.whatwg.org/#fetch-controller-state as an analogue.

annevk (approved):
Thanks, I agree with @inexorabletash. We do want to format strings the same way as enums, but I think we're doing that already.

PR #136, opened 2023-06-21
Meta: Fix broken link in FileSystemObserver proposal

a-sully:
Also clarifies one sentence about resource consumption on Linux. File descriptors and watch descriptors are each consumed (though generally more of the latter)


inexorabletash (approved):
LGTM.

PR #139, opened 2023-07-06
Support locking of directory entries

### LINKED ISSUE #137: Support locking of directory entries
Currently, a [lock](https://fs.spec.whatwg.org/#file-entry-lock) may only be taken on a file entry. This should be extended to support locking directories, too.

This is one of the [proposed changes](https://github.com/whatwg/fs/blob/main/proposals/MultipleReadersWriters.md#preventing-modification-of-parents) to support #34, but it's also needed to properly support operations which modify a directory (such as `remove()` #9, `removeEntry()`, and `move()` #10 of directories)

Operations which modify a directory should require taking an exclusive lock on the directory, which should only succeed if no contained file is locked

nathanmemmott:
Moves the locking algorithm from file entry to file system entry.

Fixes #137

   * Chrome
   * …
   * …
   * Chromium: https://crbug.com/1382215
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/fs/139.html" title="Last updated on Oct 30, 2023, 11:25 PM UTC (e26eac0)">Preview</a> | <a href="https://whatpr.org/fs/139/dd99f3a...e26eac0.html" title="Last updated on Oct 30, 2023, 11:25 PM UTC (e26eac0)">Diff</a>


index.bs
```diff
@@ -123,22 +125,16 @@ Issue: We should consider having further normative restrictions on file names th
...
 systems.
 
-A <dfn export id=file>file entry</dfn> additionally consists of
-<dfn for="file entry" export>binary data</dfn> (a [=byte sequence=]), a
-<dfn for="file entry">modification timestamp</dfn> (a number representing the number of milliseconds since the <a spec=FileAPI>Unix Epoch</a>),
-a <dfn for="file entry">lock</dfn> (a string that may exclusively be "`open`", "`taken-exclusive`" or "`taken-shared`")
-and a <dfn for="file entry">shared lock count</dfn> (a number representing the number shared locks that are taken at a given point in time).
-
-A user agent has an associated <dfn>file system queue</dfn> which is the
+A user agent has an associated <dfn>file system queue</dfn>  which is the
```
a-sully, line 128:
```suggestion
A user agent has an associated <dfn>file system queue</dfn> which is the
```
Remove (accidentally?) added space

index.bs
```diff
@@ -123,22 +125,16 @@ Issue: We should consider having further normative restrictions on file names th
...
 result of [=starting a new parallel queue=]. This queue is to be used for all
 file sytem operations.
 
 <div algorithm>
-To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of
-"`exclusive`" or "`shared`" on a given [=file entry=] |file|:
+To <dfn for="file system entry/lock">take</dfn> a [=file system entry/lock=] with a |value| of
```
a-sully, line 133: FWIW it would be nice to reword this definition to make lock take/release more clear (as was earlier suggested here  https://github.com/whatwg/fs/pull/95#discussion_r1106855972). For the sake of keeping the diffs clean, let's do this in a follow-up?

a-sully (approved):
LGTM % nit

index.bs
```diff
@@ -109,7 +109,9 @@ specifications do not need to bother implementing [=/file system entry=]'s
 
 Issue(101): Make access check algorithms associated with a FileSystemHandle.
 
-Each [=/file system entry=] has an associated <dfn for="file system entry" id=entry-name>name</dfn> (a [=string=]).
+Each [=/file system entry=] has an associated <dfn for="file system entry" id=entry-name>name</dfn> (a [=string=]),
```
a-sully, line 112:
```suggestion
Each [=/file system entry=] has an associated <dfn for="file system entry" id=entry-name>name</dfn> (a [=string=]),
```
Remove trailing whitespace

index.bs
```diff
@@ -123,22 +125,16 @@ Issue: We should consider having further normative restrictions on file names th
...
 <div algorithm>
-To <dfn for="file entry/lock">take</dfn> a [=file entry/lock=] with a |value| of
-"`exclusive`" or "`shared`" on a given [=file entry=] |file|:
+To <dfn for="file system entry/lock">take</dfn> a [=file system entry/lock=] with a |value| of
+"`exclusive`" or "`shared`" on a given [=/file system entry=] |entry|:
 
-1. Let |lock| be the |file|'s [=file entry/lock=].
-1. Let |count| be the |file|'s [=file entry/shared lock count=].
+1. Let |lock| be the |entry|'s [=file system entry/lock=].
+1. Let |count| be the |entry|'s [=file system entry/shared lock count=].
```
a-sully, line 137: This doesn't take into account locked children. Will that be done in a separate PR?

nathanmemmott: I can do it in a follow up PR. Does it make sense to do it that way?

a-sully: For now this PR is just a mechanical change to move locking from _file entry_ -> _file system entry_, but as written, it doesn't fully fix #137. Personally, I'd prefer to account for the locked children here, but if you'd prefer we can land this in two steps

If so, we should update the PR description to better describe that this is just a mechanical change, and add something like `Issue(137): Taking a lock on a directory should only succeed if no contained file is locked.` in the algorithm

index.bs
```diff
@@ -153,6 +156,24 @@ Note: These steps have to be run on the [=file system queue=].
...
+
+1. Let |lock| be the |entry|'s [=file system entry/lock=].
+1. If |lock| is not "`open`":
+   1. Return true.
+1. If |entry| is a [=directory entry=]:
+   1. [=set/For each=] |child| of |entry|'s [=directory entry/children=]:
+      1. Let |hasDescendantLock| be the result of [=file system entry/lock/checking for a descendant lock=] on |child|.
+      1. If |hasDescendantLock| is true:
+         1. Return true.
+1. Return false.
```
a-sully, line 171: The name of this algorithm doesn't give any indication that this returns something, or what the return value means. It could be nice to return something that's more self-explanatory

 For example, the "take" algorithm above returns "`failure`" or "`success`". Maybe this should return "`taken`" or "`open`"?

index.bs
```diff
@@ -153,6 +156,24 @@ Note: These steps have to be run on the [=file system queue=].
 
 </div>
 
+<div algorithm>
+To <dfn for="file system entry/lock">check for a descendant lock</dfn> on a given
+[=/file system entry=] |entry|:
+
+1. Let |lock| be the |entry|'s [=file system entry/lock=].
+1. If |lock| is not "`open`":
+   1. Return true.
```
a-sully, line 165: Currently, "check for a descendant lock" also checks whether _this_ entry is locked. IIUC this would make the "take" algorithm return early if attempting to take a "`shared`" lock on a directory entry, which doesn't seem right?

index.bs
a-sully: ... though if we're adding a new method, maybe we should just fix things in this PR rather than add a new method that follows the wrong pattern? Sorry for the confusion :)

index.bs
```diff
@@ -153,6 +156,24 @@ Note: These steps have to be run on the [=file system queue=].
...
+1. If |entry| is a [=directory entry=]:
+   1. [=set/For each=] |child| of |entry|'s [=directory entry/children=]:
+      1. Let |hasDescendantLock| be the result of [=file system entry/lock/checking for a descendant lock=] on |child|.
+      1. If |hasDescendantLock| is true:
+         1. Return true.
```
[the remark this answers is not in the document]
a-sully: Oh sorry, I misread this. Disregard! :)

index.bs
a-sully: My point is that an algorithm called "check for a descendant lock" shouldn't care whether or not you should be able to take a shared lock on a directory. The fact that it does is a big gotcha IMHO

nathanmemmott: Ah I see good point!

index.bs
nathanmemmott: Does this look good?

a-sully: Sorry for not clarifying earlier, but I believe this should be `<dfn for="file system entry">take a lock</dfn>`

Spec language maps roughly to object-oriented programming, in that `[=file system entry/lock/take=]` suggests `FileSystemLock.Take()`, while `[=file system entry/take a lock=]` suggest `FileSystemEntry.TakeALock()`. The latter is more appropriate here since we're passing in a "file system entry" to this algorithm, not a lock. That confusion is what lead to https://github.com/whatwg/fs/pull/95#discussion_r1106855972 (assuming I'm reading that correctly)

index.bs
```diff
@@ -125,16 +125,20 @@ Issue: We should consider having further normative restrictions on file names th
...
 file sytem operations.
 
 <div algorithm>
-To <dfn for="file system entry/lock">take</dfn> a [=file system entry/lock=] with a |value| of
+To <dfn for="file system entry">take a lock</dfn> with a |value| of
 "`exclusive`" or "`shared`" on a given [=/file system entry=] |entry|:
 
 1. Let |lock| be the |entry|'s [=file system entry/lock=].
 1. Let |count| be the |entry|'s [=file system entry/shared lock count=].
+1. If |entry|  is a [=directory entry=]:
```
a-sully, line 138:
```suggestion
1. If |entry| is a [=directory entry=]:
```

index.bs
```diff
@@ -153,6 +157,24 @@ Note: These steps have to be run on the [=file system queue=].
 
 </div>
 
+<div algorithm>
+To <dfn for="file system entry/lock">check for a descendant lock</dfn> on a given
```
a-sully, line 161:
```suggestion
To <dfn for="file system entry">check for a descendant lock</dfn> on a given
```
and update callers

a-sully (approved):
LGTM with some nits. Thanks for persevering on this change!

I'll stamp my approval, but given that this includes some new behaviors (mainly, that taking a lock on a directory can only succeed if no children are locked) I'd like to get a +1 from someone from another browser vendor. @annevk care to take a look?

Let me know if you have more questions about the rationale for this behavior or its implications. It's discussed briefly [here](https://github.com/whatwg/fs/blob/main/proposals/MultipleReadersWriters.md#preventing-modification-of-parents) and [here](https://github.com/whatwg/fs/blob/main/proposals/MultipleReadersWriters.md#not-locking-parents). The perhaps unexpected (if you're familiar with Posix, at least) implications are that moving or deleting a directory that has an open `FileSystemSyncAccessHandle` will fail

Both of these operations "just work" on Posix, but not on other platforms (for example, on Windows you can't delete a directory while there's an open file handle). With the web being a cross-platform platform, I don't think we should be exposing implementation details of Posix

(we should probably add a note in the spec about this...)

nathanmemmott:
@jesup @szewai @annevk Have you had a chance to take a look at this?

index.bs
```diff
@@ -159,11 +159,29 @@ Note: These steps have to be run on the [=file system queue=].
...
+    1. Return "`taken`".
+  1. If |child| is a [=directory entry=]:
+    1. Let |descendantLockStatus| be the result of [=file system entry/checking for a descendant lock=] on |child|.
+    1. If |descendantLockStatus| is "`taken`":
+      1. Return "`taken`".
+1. Return "`open`".
 
-1. Let |lock| be the |file|'s associated [=file entry/lock=].
-1. Let |count| be the |file|'s [=file entry/shared lock count=].
+Note: These steps have to be run on the [=file system queue=].
```
annevk, line 175: This makes more sense to note at the bottom of "take a lock" as that's the one that will be invoked directly from elsewhere, no?

annevk (commented):
This seems reasonable to me, but I'm not entirely comfortable signing off for WebKit. I'd like at least @szewai to weigh in as well.

szewai:
> LGTM with some nits. Thanks for persevering on this change!
> 
> I'll stamp my approval, but given that this includes some new behaviors (mainly, that taking a lock on a directory can only succeed if no children are locked) I'd like to get a +1 from someone from another browser vendor. @annevk care to take a look?
> 
> Let me know if you have more questions about the rationale for this behavior or its implications. It's discussed briefly [here](https://github.com/whatwg/fs/blob/main/proposals/MultipleReadersWriters.md#preventing-modification-of-parents) and [here](https://github.com/whatwg/fs/blob/main/proposals/MultipleReadersWriters.md#not-locking-parents). The perhaps unexpected (if you're familiar with Posix, at least) implications are that moving or deleting a directory that has an open `FileSystemSyncAccessHandle` will fail
> 
> Both of these operations "just work" on Posix, but not on other platforms (for example, on Windows you can't delete a directory while there's an open file handle). With the web being a cross-platform platform, I don't think we should be exposing implementation details of Posix
> 
> (we should probably add a note in the spec about this...)

(Sorry if I have missed any information)  I don't quite understand the rationale of declaring lock on a directory entry. In what cases will we need to take / release the lock? If we just want to ensure no children is locked when directory is removed or moved, we could just perform the check directly before the operation. 

I could understand a directory entry might have a "isLocked" state to indicate whether any of its children holds lock, but not sure about a directory itself has a lock.

nathanmemmott:
Mistakenly closed while cleaning up branches.

index.bs
nathanmemmott: This is noted at the bottom of "take a lock". Should it only be noted there?

nathanmemmott:
> > LGTM with some nits. Thanks for persevering on this change!
> > I'll stamp my approval, but given that this includes some new behaviors (mainly, that taking a lock on a directory can only succeed if no children are locked) I'd like to get a +1 from someone from another browser vendor. @annevk care to take a look?
> > Let me know if you have more questions about the rationale for this behavior or its implications. It's discussed briefly [here](https://github.com/whatwg/fs/blob/main/proposals/MultipleReadersWriters.md#preventing-modification-of-parents) and [here](https://github.com/whatwg/fs/blob/main/proposals/MultipleReadersWriters.md#not-locking-parents). The perhaps unexpected (if you're familiar with Posix, at least) implications are that moving or deleting a directory that has an open `FileSystemSyncAccessHandle` will fail
> > Both of these operations "just work" on Posix, but not on other platforms (for example, on Windows you can't delete a directory while there's an open file handle). With the web being a cross-platform platform, I don't think we should be exposing implementation details of Posix
> > (we should probably add a note in the spec about this...)
> 
> (Sorry if I have missed any information) I don't quite understand the rationale of declaring lock on a directory entry. In what cases will we need to take / release the lock? If we just want to ensure no children is locked when directory is removed or moved, we could just perform the check directly before the operation.
> 
> I could understand a directory entry might have a "isLocked" state to indicate whether any of its children holds lock, but not sure about a directory itself has a lock.

Yes this is just for move and remove for right now. Move and remove will be available for both files and directories. We'll need to take some sort of exclusive lock during these operations whether that's a `file system entry/lock` or `isLocked`. The `file system entry/lock` is reused to simplify locking logic. For example, `isLocked` would have to be added to `file system entry/take a lock` steps as an alternative way to exclusively lock.

nathanmemmott:
Latest push adds a missing check for ancestor locks.

Issue #17, opened 2022-03-09
Interactions with the BFCache


a-sully:
_Migrated from https://github.com/WICG/file-system-access/issues/319_ (see https://github.com/whatwg/fs/issues/2)

### @ArthurSonzogni

You may want to reference this from the BFCache meta bug:
https://github.com/whatwg/html/issues/5880
+CC @rakina

I was wondering what kind of interaction this feature will have with the Back-forward cache. I see the execution context can obtain a lock for a file. Then the document might be (?) able to enter the BackForwardCache without releasing it. Maybe this could cause some problems?
For instance there are other features like WebLock which prevents the BFCache from being used. Maybe the same should happen here:
https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/back_forward_cache_impl.cc;l=265;drc=e5616aeff413a17175a96056b3bf1fbc9ca0ade7;bpv=1;bpt=1
Maybe you did it already. I see kWebFilesystem. Not sure if this is related.

More generally, you might want to clarify if there are any BFCache specific behavior about this feature.

### @rakina

On the kWebFilesystem thing, I think @hajimehoshi and @fergald did add WebFileSystem support for BFCache, but I'm not sure about the details - maybe the can comment on what Chrome is doing and if there's anything to be added on the spec side.

### @fergald

My understanding is that WebFileSystem had nothing that prevented bfcaching (there are no connections to bring up/down or locks). I see that it's unblocking is still controlled by a flag but we should clean that up and leave it unconditionally unblocked.

I don't think there's any spec change (unless we want to state positively that it is compatible with bfcache) but we should probably add a WPT that checks if a page using WFS is cached. That would make it clear which browsers support it.

### @mkruisselbrink 

On the chrome side currently using the file system access API with local files on the file system will cause the page to be prevented from entering bfcache, although I think that is primarily since we haven't reased about interaction with usage tracking etc for our current permissions implementation (where lifetime of permissions is dependent on the presence of tabs with pages loaded).

For the newly proposed access handle API, which include file locking, I do suspect we'll need explicit spec-level integration with bfcache. We don't want pages in the bfcache to be able to keep files locked after all.

rakina:
Thanks! Wondering if there's any resolution needed here. BTW, there's a [guide](https://w3ctag.github.io/design-principles/#support-non-fully-active) on how to handle non-fully active documents (which includes bfcached documents) in spec.

a-sully:
[...]

PR #142, opened 2023-08-14
Add BFCache section to multi reader writer proposal

nathanmemmott:
Adds a section to the proposal to explain interaction with the BFCache.

***
<a href="https://whatpr.org/fs/142.html" title="Last updated on Dec 9, 2023, 1:02 AM UTC (5397fe3)">Preview</a> | <a href="https://whatpr.org/fs/142/bd6ef5a...5397fe3.html" title="Last updated on Dec 9, 2023, 1:02 AM UTC (5397fe3)">Diff</a>

### DIFF
--- a/proposals/MultipleReadersWriters.md
+++ b/proposals/MultipleReadersWriters.md
@@ -225,6 +225,11 @@ childAccessHandle.close();
 await parentHandle.remove(); // successful
 ```
 
+### Interactions with BFCache
+A page may still hold a file system lock when it enters the BFCache. A fully active page could then be made aware of a BFCached page if there is contention between locks they hold.
+
+To keep BFCache enabled when a site uses the File System Access API, a BFCached page must be evicted on locking contention with a fully active page (whether or not it is of the same origin). Otherwise, a file system lock held by a page will not affect the page's eligibility for BFCache. This allows the site to have the performance gains of BFCache up until it would be made aware of the BFCache.
+
 ## Alternatives Considered
 
 ### Not Locking File Entry


proposals/MultipleReadersWriters.md
```diff
@@ -221,6 +221,11 @@ childAccessHandle.close();
 await parentHandle.remove(); // successful
 ```
 
+### Interactions with BFCache
+A page should not be made aware that it or another page is or has been in the BFCache. A fully active page could be made aware of a BFCached page if there is contention between locks they hold. The File System Access API operations/primitives hold locks that may cause this issue.
```
a-sully, line 225: The key issue is that _a page may still hold a file system lock when it enters the BFCache_. The last sentence hints at this, but I think this paragraph could be more concise if that was stated more directly

proposals/MultipleReadersWriters.md
```diff
@@ -221,6 +221,11 @@ childAccessHandle.close();
...
+
+To keep BFCache enabled when a site uses the File System Access API, a BFCached page should be evicted only on locking contention with a fully active page. This allows the site to have the performance gains of BFCache up until it would be made aware of the BFCache.
```
a-sully, line 227: > a BFCached page should be evicted only on locking contention with a fully active page

A BFCached page may be evicted for any number of reasons unrelated to this API. Rather than say that a page will _only_ be evicted due to lock contention (which implies that no lock contention == no eviction), we should say that a page _must_ be evicted if it causes lock contention (which says nothing about what happens when there's no lock contention)

a-sully (changes requested):

proposals/MultipleReadersWriters.md
nathanmemmott: Ok updated

proposals/MultipleReadersWriters.md
```diff
@@ -221,6 +221,11 @@ childAccessHandle.close();
 await parentHandle.remove(); // successful
 ```
 
+### Interactions with BFCache
+The File System Access API operations/primitives hold locks in the BFCache. A page should not be made aware that it or another page is or has been in the BFCache. A fully active page could be made aware of a BFCached page if there is contention between locks they hold.
+
+To keep BFCache enabled when a site uses the File System Access API, a BFCached page must be evicted on locking contention with a fully active page. Otherwise, a file system lock held by a page will not affect the page's eligibility for BFCache. This allows the site to have the performance gains of BFCache up until it would be made aware of the BFCache.
```
a-sully, line 227: > This allows the site to have the performance gains of BFCache up until it would be made aware of the BFCache.

This sentence is a bit misleading, since it suggests there is just "the site", whereas there are two pages in question here: (1) the page in the BFCache, and (2) the active page. These pages may not correspond to the same site for local file system access

proposals/MultipleReadersWriters.md
```diff
@@ -221,6 +221,11 @@ childAccessHandle.close();
 await parentHandle.remove(); // successful
 ```
 
+### Interactions with BFCache
+The File System Access API operations/primitives hold locks in the BFCache. A page should not be made aware that it or another page is or has been in the BFCache. A fully active page could be made aware of a BFCached page if there is contention between locks they hold.
```
a-sully, line 225:
```suggestion
A page may still hold a file system lock when it enters the BFCache. A fully active page could then be made aware of a BFCached page if there is contention between locks they hold.
```

a-sully (approved):
LGTM with comments addressed

a-sully:
Let's update the spec text itself in a follow-up (and address https://github.com/whatwg/fs/issues/17 in doing so)

PR #163, opened 2024-06-27
Update FileSystemObserver explainer

dslee414:
- Updates `FileSystemChangeType` names from created/deleted/unsupported to appeared/disappeared/unknown:
  - When a file or directory has been moved into (or out of) the watched scope, the file/dir is to be reported as "created" or "deleted" with respect to the scope; however, this naming may be confusing as it is not really newly created file/dir (or deleted). Therefore, new names "appeared" and "disappeared" are suggested instead.
  - "unsupported" is renamed to "unknown" to indicate a general use case where a type is not known or not to be specified to the site for various reasons.
- Change the data type of `FileSystemChangeRecord.relativePathMovedFrom`
  - Providing a `FileSystemHandle` to a non-existing file would not make much sense, but it would be still useful for sites to know where a file/dir has been moved from. It should return an array of path components instead (similar to `relativePathComponents`).
- Update stakeholder signal

### DIFF
--- a/proposals/FileSystemObserver.md
+++ b/proposals/FileSystemObserver.md
@@ -304,6 +304,10 @@ Likewise, changes which occur before an observer is created should not be report
 
 A `FileSystemObserver` is not [serializable](https://html.spec.whatwg.org/multipage/structured-data.html#serializable) and therefore cannot be persisted across browsing sessions. Websites which wish to watch the same files on each session may store serializable `FileSystemHandle` and `FileSystemObserverObserveOptions` objects in IndexedDB, then create a `FileSystemObserver` and configure it from these objects on page reload.
 
+### Interactions with Back/forward Cache
+
+If changes occurred while the page was not fully active, and the page becomes active again (i.e. back/forward cache), then user agents may use the `"unknown"` `FileSystemChangeType` to indicate that _changes_ have occurred. Specific types and ordering of changes should not be exposed but indicating that some changes have occurred could be useful to the website to perform any special handling.
+
 ### Signaling Changes Made via a `FileSystemSyncAccessHandle`
 
 It is assumed that a user agent’s implementation of the `FileSystemObserver` interface will involve coordinating with a centralized browser process. However, unlike most web storage APIs, reading and writing files with a `FileSystemSyncAccessHandle` is commonly implemented largely without coordinating with a centralized browser process. This is critical to the exceptional performance characteristics of this interface. `write()` or `truncate()` operations on a `FileSystemSyncAccessHandle` should trigger a file system change record, but requiring round-trip IPC to complete before synchronously returning would be detrimental to performance.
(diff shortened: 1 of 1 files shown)


proposals/FileSystemObserver.md
```diff
@@ -374,7 +374,7 @@ await observer.observe(directoryHandle, { recursive: true });
 // corresponding to `directoryHandle`, then `touch file.txt`)
 ```
 
-User agents may also use the `"unsupported"` `FileSystemChangeType` to explicitly indicate that change types are not supported.
+User agents may also use the `"unknown"` `FileSystemChangeType` to indicate that a change has occurred but the type of change is unknown or is not to be specified to the site. This could mean that the change type does not belong to any other values defined in `FileSystemChangeType` due to cross-platform differences, or a change has occurred while the site was not fully active (i.e. site in back/forward cache becomes fully active again).
```
nathanmemmott, line 377: I really don't think you should say it could be a change type that doesn't belong to any other values. Currently unknown is being used to say it "unknown" if zero or more events have occurred, so that the developer will trigger a poll.

Sending out "unknown" for OS events that are specific to some platform would be a confusing signal to developers.

dslee414: Rephrased this section into a new section about interactions with BFCache, as it is not related to cross-platform differences.

proposals/FileSystemObserver.md
```diff
@@ -219,8 +219,8 @@ async function markDirty(record) {
       markMoved(record.root, record.relativePathMovedFrom,
                 record.relativePathComponents);
       break;
-    case 'unsupported':
-      // Change types may not be supported on all platforms.
+    case 'unknown':
+      // Change type is not known.
```
nathanmemmott, line 223: Could this be updated to zero or more changes may have been missed?

dslee414: It can't be zero, can it? In any case, updated this comment to indicate the possibility of multiple events

nathanmemmott: Yeah it can. On Windows, if the scope of the OS watch is larger than the observation's scope, then it can buffer overflow for file events outside the observation's scope. But we can't know that, and have to assume that there could have been events in the observation's scope, so we'll have to send an unknown. There could be a similar scenarios on Mac.

a-sully: FWIW this is just an explainer (not a spec) so I don't think it's necessary to enumerate all the possible scenarios the implementation may choose to use this change type. This document already has several hedges about how changes are "best-effort" etc., so this seems fine to me ¯\\_(ツ)_/¯

a-sully (approved):

a-sully:
Please ensure to follow the guidelines at https://github.com/whatwg/meta/blob/main/COMMITTING.md when you submit this PR. Thanks!

annevk:
It seems the guidelines were not followed. :-( This should have been a "Meta: " change as it doesn't change the standard. The commit body also repeats the commit title and the commit title contains a redundant reference to the PR...

dslee414:
Apologies! I misunderstood what "Meta: " prefix is for.

PR #164, opened 2024-06-28
Editorial: Define the File System concept

nathanmemmott:
Pulls out some implementation-defined concepts into a file system concept.

This is so that the file system concept can be extended further in the future.


a-sully:
Discussed offline. Right now this change isn't really editorial, since it introduces a new "file system" definition, among other changes. I suggest splitting the PR into (at least) two:
1) An editorial reorganization (i.e. no new dfns; just refactoring)
2) Substantiative changes, such as introducing the new "file system" definition, etc

PR #165, opened 2024-07-18
File system observer

nathanmemmott:
Add the FileSystemObserver API

Defines the FileSystemObserver API, an API to observe file system
change events.

See #123 and WICG/file-system-access#72

   * Chrome
   * [Firefox](https://github.com/mozilla/standards-positions/issues/942)
   * … 
   * Chromium: https://crbug.com/320560741
   * Gecko: …
   * WebKit: …
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 


index.bs
```diff
@@ -55,13 +55,326 @@ before a user has picked a location to save to, without forcing the website to u
...
+     |changedHandle|'s [=FileSystemHandle/locator=] is equal to
+     |observationLocator|:
+      1. Set |errorRecord| to the result of
+        <a>creating a new `FileSystemChangeRecord`</a> for |observation| given
+        |changedHandle|, "{{FileSystemChangeType/errored}}", and null.
+      1. [=list/Append=] |errorRecord| to |records|.
+      1. [=file system observation/Destroy observation=] |observation|.
+      1. [=break=].
+
+1. [=file system observation/Invoke the callback=] of the |observation| with
```
TomokiMiyauci, line 328: If records is empty, invoke should not be called.

index.bs
```diff
@@ -249,69 +530,98 @@ A <dfn export>file locator</dfn> is a [=/file system locator=] whose
...
+Note: These steps have to be run on the [=file system queue=].
 
 </div>
 
 <div algorithm>
-The <dfn export data-lt="getting the locator">get the locator</dfn> algorithm given
-[=/file system entry=] |entry| runs an [=implementation-defined=] series of steps adhering to these
-constraints:
+To <dfn for="file system locator">get the relationship</dfn> between a
+[=/file system locator=] |self| to a [=file system locator=] |related|:
```
mkruisselbrink, line 573: nit: "the relationship between a *and* b" or "the relationship *from* a to b" rather than "between a to b"?

index.bs
```diff
@@ -249,69 +530,98 @@ A <dfn export>file locator</dfn> is a [=/file system locator=] whose
...
 
-<p class=example id=example-path>Let |path| be the [=/list=]
-« "`data`", "`drafts`", "`example.txt`" ».
-There is no expectation that a file named `example.txt` exists anywhere on disk.
+1. Let |fileSystem| be |locator|'s [=file system locator/file system=].
+1. Let |path| be |locator|'s [=file system locator/path=].
+1. Let |entry| be the result of running |fileSystem|'s
+   [=file system/locate an entry=] given |path|.
+1. If |entry| is null, return null.
+1. If |locator| is a [=file locator=], [=Assert=]: |entry| is a [=file entry=].
```
mkruisselbrink, line 605: Are these "Assert"s actually valid as assertions? As far as I can tell it is totally possible for this to be called with a locator whose type doesn't match the entry that exists now (for example if you have a FileHandle with a certain locator, delete the entry, re-create it as a directory, and then try to do anything with the file handle).

index.bs
```diff
@@ -1085,10 +1409,15 @@ if (relative_path === null) {
...
-The <dfn method for=FileSystemDirectoryHandle>resolve(|possibleDescendant|)</dfn> method steps are
-to return the result of [=file system locator/resolving=]
-|possibleDescendant|'s [=FileSystemHandle/locator=]
-relative to [=this=]'s [=FileSystemHandle/locator=].
+The <dfn method for=FileSystemDirectoryHandle>resolve(|possibleDescendant|)</dfn>
+method steps are:
+
+1. Let |result| be [=a new promise=].
+1. [=Enqueue the following steps=] to the [=file system queue=]:
+  1. [=/resolve=] |result| with the result of [=file system locator/resolving=]
```
mkruisselbrink, line 1417: You can't call "/resolve" on a parallel queue. You need to post a task back to the correct event loop to resolve the actual promise.

index.bs
```diff
@@ -55,13 +55,326 @@ before a user has picked a location to save to, without forcing the website to u
...
+
 
 # Files and Directories # {#files-and-directories}
 
 ## Concepts ## {#concepts}
 
+### File System ### {#concept-file-system}
+
+A <dfn export>file system</dfn> is an [=implementation-defined=]
+[=storage endpoint=] that maintains a mapping of [=file system path=]s to
```
mkruisselbrink, line 72: I don't think it makes sense for a "file system" to be a storage endpoint. A storage endpoint is defined as "an API" after all. The bucket file system somehow should still be a storage endpoint, but a generic file system concept here I don't think should have anything to do with a "storage endpoint".

index.bs
```diff
@@ -1736,6 +2067,192 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+  constructor(FileSystemObserverCallback callback);
+
+  Promise<undefined> observe(FileSystemHandle handle,
+                          optional FileSystemObserverObserveOptions options = {});
+  undefined unobserve(FileSystemHandle handle);
+  undefined disconnect();
+};
+</xmp>
+
+The {{FileSystemObserver}} interface can be used to observe
```
mkruisselbrink, line 2097: nit: this should possibly be a non-normative ("for developers") note

index.bs
```diff
@@ -1736,6 +2067,192 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+  "modified",
+  "moved",
+  "unknown",
+};
+
+[
+    Exposed=(DedicatedWorker,SharedWorker,Window),
+    SecureContext,
+    RuntimeEnabled=FileSystemObserver
+] interface FileSystemChangeRecord {
```
mkruisselbrink, line 2202: I wonder why this is an interface? It seems this could just as well be a dictionary, which might make some things simpler? If this does have to be an interface for some reason, it at least needs a constructor per https://www.w3.org/TR/design-principles/#constructors.

index.bs
```diff
@@ -1736,6 +2067,192 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+</xmp>
+
+<div dfn-type="attribute" dfn-for="FileSystemChangeRecord">
+  : <dfn>root</dfn>
+  :: The handle that was passed to FileSystemObserver.observe().
+  : <dfn>changedHandle</dfn>
+  :: The path of `changedHandle` relative to `root`.
+  : <dfn>type</dfn>
+  :: The type of change.
+  : <dfn>relativePathMovedFrom</dfn>
```
mkruisselbrink, line 2218: nit: add a definition for relativePathComponents as well? Although all these read more like non-normative web developer documentation (which is also good to have) rather than normative definitions.

mkruisselbrink (commented):
It would be nice if we could somehow specify more precisely what exact change events are triggered by changes to the bucket file system, since that should be entirely under control of user agents, so should be possible to be fully compatible between implementations. Not sure how to be go about doing that. Left a bunch of other comments as well...

index.bs
```diff
@@ -55,13 +55,326 @@ before a user has picked a location to save to, without forcing the website to u
...
+Note: The [=file system/file system event/entry type=] is null when the
+[=/file system=] cannot determine the {{FileSystemHandleKind}} of the
+[=file system/file system event/modified path=] due to the underlying
+[=file system entry=] no longer being there.
+
+A [=/file system=] has an associated [=/set=] of
+[=file system observations=] <dfn for="file system">observations</dfn>
+which is initialized to the empty set.
+
+When a [=/file system=] |fileSystem| fires a [=/list=] of
```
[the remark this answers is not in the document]
nathanmemmott: Yeah I tried to set this change up so that we can eventually spec out how the BFS sends file system change events. Let me know if you want me to explore that in this PR.

nathanmemmott:
The errors causing the build to fail were not introduced by this PR.

index.bs
```diff
@@ -1721,13 +2066,10 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
   1. Set |dir|'s [=directory entry/children=] to an empty [=/set=].
   1. Set |map|["root"] to |dir|.
 
-1. Let |root| be an [=implementation-defined=] opaque [=string=].
+1. Let |fileSystem| be [=/bucket file system=]'s [=file system/root=].
```
mkruisselbrink, line 2069: Somewhere in here fileSystem needs to be linked to a specific storage bottle for the correct storage bucket/origin. As written now it reads like there is one shared file system that all the storage buckets access. The old spec didn't really do this rigorously either, so it's probably okay to keep it a little bit hand-wavy (i.e. you might not need to go the whole way towards calling "obtain a local storage bottle map", and making it explicit that all the data for the file system is stored in the returned map), but we certainly don't want the spec to read like there is one "the bucket file system". Each storage bucket does need its own file system.

index.bs
```diff
@@ -1736,6 +2078,227 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+
+  1. [=Queue a storage task=] with |global| to run these steps:
+    1. If |accessResult|'s [=file system access result/permission state=]
+       is not "{{PermissionState/granted}}", [=/reject=] |result| with a
+       {{DOMException}} of |accessResult|'s
+       [=file system access result/error name=] and abort these steps.
+
+    1. If |entry| is null, [=/reject=] |result| with a
+       "{{NotFoundError}}" {{DOMException}} and abort these steps.
+    1. [=Assert=]: |entry| is a [=file entry=].
```
mkruisselbrink, line 2161: I don't understand both what makes this a valid assertion (nothing above seems to guarantee this), or why we would want/need this invariant; don't we want to allow observing directories?

index.bs
```diff
@@ -1736,6 +2078,227 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+  "moved",
+  "unknown",
+};
+
+[
+    Exposed=(DedicatedWorker,SharedWorker,Window),
+    SecureContext,
+    RuntimeEnabled=FileSystemObserver
+] interface FileSystemChangeRecord {
+  constructor(FileSystemHandle root,
```
mkruisselbrink, line 2219: Rather than having individual parameters for each attribute, the constructor should take a FileSystemChangeRecordInit dictionary, which has members matching those of the interface.

index.bs
nathanmemmott: Let me know if the update makes sense.

index.bs
```diff
@@ -1736,6 +2079,230 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+  required sequence<USVString> relativePathComponents;
+  required FileSystemChangeType type;
+  sequence<USVString>? relativePathMovedFrom;
+};
+
+[
+    Exposed=(DedicatedWorker,SharedWorker,Window),
+    SecureContext,
+    RuntimeEnabled=FileSystemObserver
+] interface FileSystemChangeRecord {
```
annevk, line 2226: Is there a reason this is a class and not just a dictionary? I think in retrospect DOM's `MutationRecord` could have just been a dictionary as well. It's also not entirely clear to me why this has a constructor.

index.bs
```diff
@@ -1736,6 +2079,230 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+  boolean recursive = false;
+};
+
+callback FileSystemObserverCallback = undefined (
+  sequence<FileSystemChangeRecord> records,
+  FileSystemObserver observer
+);
+
+[
+    Exposed=(DedicatedWorker,SharedWorker,Window),
```
youennf, line 2097: Why not `(Worker, Window)`, `FileSystemHandle` is probably exposed in service worker contexts too?

index.bs
```diff
@@ -1736,6 +2079,230 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+    1. [=file system observer registration/Destroy=] |observerRegistration|.
+
+</div>
+
+## The {{FileSystemChangeRecord}} interface ## {#api-filesystemchangerecord}
+
+<xmp class=idl>
+enum FileSystemChangeType {
+  "appeared",
+  "disappeared",
```
youennf, line 2207: We usually talk about creating and removing or deleting files/folders. I guess these two cover these cases. Why this choice of name?
It might be good to have a definition for each of these values.

index.bs
```diff
@@ -1736,6 +2079,230 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+  "errored",
+  "modified",
+  "moved",
+  "unknown",
```
youennf, line 2211: I am not sure when `errored` and `unknown` would happen.
I guess `errored` could happen if unmounting a disk for instance.
I have no clue for `unknown`.
If we do not have a good usecase, should we remove from the initial PR?
Giving a description explaining what these values are might be nice.

index.bs
```diff
@@ -1736,6 +2079,230 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+</div>
+
+## The {{FileSystemChangeRecord}} interface ## {#api-filesystemchangerecord}
+
+<xmp class=idl>
+enum FileSystemChangeType {
+  "appeared",
+  "disappeared",
+  "errored",
+  "modified",
```
youennf, line 2209: What is `modified` meaning?
If only touching the file, will it trigger a FileSystemChange?
If so, I would guess that this could trigger events where no information for the web page has changed (IIRC, access/modification dates are not exposed).
Or maybe it is restricted to changes in the content itself?

index.bs
```diff
@@ -1736,6 +2079,230 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
 
 </div>
 
+# Observing file system change events # {#observing-the-file-system}
+
+## The {{FileSystemObserver}} interface ## {#api-filesystemobserver}
+
+<xmp class=idl>
+dictionary FileSystemObserverObserveOptions {
+  boolean recursive = false;
```
youennf, line 2088: If a folder being observed is removed, which callback will be called?
One for the folder, or one per each file in the folder, or depending on the OS?

index.bs
```diff
@@ -1736,6 +2079,230 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+
+## The {{FileSystemChangeRecord}} interface ## {#api-filesystemchangerecord}
+
+<xmp class=idl>
+enum FileSystemChangeType {
+  "appeared",
+  "disappeared",
+  "errored",
+  "modified",
+  "moved",
```
youennf, line 2210: Is `moved` also covering `rename` change?

index.bs
mkruisselbrink: I asked the same question, and we had some offline discussion about this. The question of dictionary vs class seems to be the oldest open tag guidance request, but the most recent thinking in the issue thread there (https://github.com/w3ctag/design-principles/issues/11#issuecomment-1515585583) seems to lean arguably in favor of using a class here, or at least not arguing against it (but that guidance certainly has changed a couple of times as the make-up of the tag has changed). I'm not sure what the latest thinking in this area is, and it would be nice if the TAG would actually write up something about this other than in that very long issue thread...

index.bs
nathanmemmott: A separate API would be needed for service workers. The service worker becomes inactive after 30 seconds of no events. The change events received in the `FileSystemObserver` callback would not count as a service worker event. There are also privacy considerations around allowing a site to observe file changes without the user being aware.

index.bs
nathanmemmott: Added definitions of each `FileSystemChangeType`.

index.bs
nathanmemmott: Added definitions for each with examples.

index.bs
nathanmemmott: Yes

index.bs
annevk: I'm not sure that accurately reflects the upthread discussion though with input from me and @domenic. In particular @domenic made the point there that for observer-related cases where it's just about data a dictionary would be more idiomatic. Having both a class and a constructor for it that takes a dictionary seems quite bloated.

index.bs
annevk: I wonder if symmetry with the DOM might be nice. That would argue for "connected" and "disconnected".

nathanmemmott: I could see replacing "errored" with "disconnected", but I don't think "connected" and "disconnected" communicate files moving in/being created and moving out/being deleted very well.

index.bs
domenic: Per https://groups.google.com/a/chromium.org/g/blink-dev/c/6oOaFmia2dc/m/gx0KgpQqBQAJ, the latest plan seems to be to use a dictionary, especially since that is easier to upgrade to an interface later. I think that's a good plan, and updating the spec PR to match that would be good.

nathanmemmott: Switch to dictionary.

nathanmemmott:
The latest patch adds some logic that we decided on recently. We no longer send the changed handle for "disappeared", "errored", and "unknown".

nathanmemmott:
@annevk @mkruisselbrink @domenic @youennf PTAL again and let me know if this is ready to merge.

@asutherland I don't have permission to add you but please take a look. If someone does have permission, please add him.

mkruisselbrink (approved):

mkruisselbrink:
This looks good to me, but it would be good to get at least one other person to sign off on this before merging.

nathanmemmott:
@annevk Ok updated the OP. Let me know if it looks correct.

annevk:
You haven't filed Gecko or WebKit bugs.

nathanmemmott:
Filed the bugs and updated the OP.

index.bs
szewai: Why is moved event useful? Can we just fire a created change with new path and a deleted event with old path?

index.bs
szewai: What actions are expected for clients to take for "unknown" or "errored"?
For the use cases in https://github.com/whatwg/fs/blob/main/proposals/FileSystemObserver.md, I can see how they may want to react to "created", "deleted" and "content-modified", but not clear about the others.

Also as https://github.com/whatwg/fs/blob/main/proposals/FileSystemObserver.md#cross-platform-compatibility mentioned, we should consider about adding types that are generally supported, and `unknown` seems like platform specific?

annevk:
@asutherland I was pointed towards https://github.com/mozilla/standards-positions/issues/942#issuecomment-2113526096. Did you mean to leave that comment here as well?

index.bs
```diff
@@ -1736,6 +2080,234 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+
+callback FileSystemObserverCallback = undefined (
+  sequence<FileSystemChangeRecord> records,
+  FileSystemObserver observer
+);
+
+[
+    Exposed=(DedicatedWorker,SharedWorker,Window),
+    SecureContext
+] interface FileSystemObserver {
```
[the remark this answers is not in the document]
szewai: Discussed with @youennf, and we think an alternative proposal could be:
1. Add a new interface `FileSystemChangeEvent` to be dispatched to `FileSystemHandle`. 
2. Add a new property `observeOptions` to `FileSystemHandle`. Client can view / configure types of event to receive (default options could be "not observing" so we don't monitor file system changes).

This should make the interfaces and garbage collection story simpler (no need to keep both `FileSystemObserver` and `FileSystemHandle` to monitor change).

index.bs
nathanmemmott: For "unknown", they may want to poll their watched directory since they may have missed events. For "errored", they may want to try watching the folder again since they'll no longer be receiving.

> unknown seems like platform specific?

An unknown can occur at least on Mac and Windows.

index.bs
nathanmemmott: We'd lose information if we reported it that way. If we receive a appeared for a/b and a disappeared for a/c, we couldn't tell if that's rename/move or if a/c was just deleted and a/b created.

index.bs
```diff
@@ -1736,6 +2080,234 @@ The <dfn method for=StorageManager>getDirectory()</dfn> method steps are:
...
+    1. [=file system observer registration/Destroy=] |observerRegistration|.
+
+</div>
+
+## The {{FileSystemChangeRecord}} interface ## {#api-filesystemchangerecord}
+
+<xmp class=idl>
+enum FileSystemChangeType {
+  "appeared",
+  "disappeared",
```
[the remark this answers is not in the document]
nathanmemmott: "appeared" and "disappeared" were chosen since these events cover "created"/"moved in to scope" and "deleted"/"moved out of scope" respectively.

index.bs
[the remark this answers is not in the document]
a-sully: I see the appeal of this approach, though there are a few things we'd have to consider. Primarily:
- What is the behavior if there exist several equivalent `FileSystemHandle`s?
- What happens when a `FileSystemHandle` is moved - does the observer move along with the handle, or not?

Let's look at a case with duplicate handles:

```
// Create two files which each point to file.txt
const file1 = dir.getFileHandle('file.txt', { create: true });
const file2 = dir.getFileHandle('file.txt');

// Start observing file1 (API TBD)

// Q: What is that state of file2.observeOptions?
```

Either:
- `file2` is "not observing"
  - This would be quite misleading. Stopping the observation would require finding _that specific_ handle you started the observation with. This seems like a non-starter, since handles are specified as independent and (mostly) interchangable pointers to paths - e.g. `FileSystemHandle.move()` only modifies _that specific handle_, not all handles for which `isSameEntry()` is true
- `file2` is "observing"
  - Does this suggest we dispatch duplicate events to `file1` and `file2`? That would be unfortunate. What about having some state to indicate a handle is "being observed from elsewhere"? We could, but that would add complexity and introduce a new pattern which is less familiar to developers than the `*Observer` pattern
  - Checking the observation status of `file2` must be async, just like `isSameEntry()`, as this requires equivalence checks with all other existing observations and/or handles

If we agree that the former option is a non-starter, then in the latter case, what happens when `file2` is moved?

```
// Move file1
await file1.move('new.txt');

// Q: What is that state of file2.observeOptions?
```

Either:
- We are now observing two paths: `file.txt` and `new.txt`. This would be very unexpected behavior IMHO
- We are no longer observing `file2`. This would require adopting the "being observed from elsewhere" pattern

---

At a high level, this relates back to https://github.com/whatwg/fs/issues/59 - i.e. is a `FileSystemHandle` is based on a file path or a reference to a file? We've since specified it is a path https://github.com/whatwg/fs/pull/96. Passing a `FileSystemHandle` to a `FileSystemObserver` indicates the path to be watched. This layer of indirection means the path being observed is stable even if we duplicate or move the original `FileSystemHandle`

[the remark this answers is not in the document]
szewai: > - What is the behavior if there exist several equivalent FileSystemHandles?

I think different `FileSystemHandle` with the same path can have different observing states. In the example above, if `file1` is observing and `file2` is not, then we only dispatch change event to `file1`. If both are observing, then we dispatch change events to both (implementer can decide whether to create 1 or 2 file system observers for the same path in this case).

> - What happens when a FileSystemHandle is moved - does the observer move along with the handle, or not?

Observing `FileSystemHandle` is observing the underlying path of it. If `FileSystemHandle` changes its underlying path, then the observer will change its path too. E.g. if `file1` (observing `file.txt`) is moved to `new.txt`, then it will start receiving change event about `new.txt`.

> Stopping the observation would require finding that specific handle you started the observation with. 

It sounds right to me the observer goes with `FileSystemHandle`. If client wants to stop observing, they may set its observing options to null and release the handle object.

> This seems like a non-starter, since handles are specified as independent and (mostly) interchangable pointers to paths - e.g. FileSystemHandle.move() only modifies that specific handle, not all handles for which isSameEntry() is true

I don't understand why this is a non-starter. The independence is exactly how client could ensure they are monitoring the path, not the file reference. E.g. if `file2` (not observing) is moved to path `new.txt`, then `file1` (observing `file.txt`) gets change event about file being deleted (or moved). If `file3` is created with path `file.txt` (`const file3 = dir.getFileHandle('file.txt', { create: true });`) then `file1` gets change event about file being created.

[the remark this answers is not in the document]
szewai: I wonder what quirks are needed for this. 

From current API proposal it's unclear that why `FileSystemObserver` is necessary. I could see a possibility to add it if the operations does not involve `FileSystemHandle`: e.g. observer is created with paths (relative path to root), and callback is invoked with path that has change. The benefit of this is client does not need to create handle (with underlying file / directory) or decide the type of the path if they only want to monitor change; and we may not need to deal with complication of which handle has changed. But the current interface requires `FileSystemHandle` anyways, so it seems adding event handler is simpler (something similar to [CookieChangeEvent](https://wicg.github.io/cookie-store/#CookieChangeEvent)), and it might be shimmed by existing `FileSystemObserver` API.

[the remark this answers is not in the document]
a-sully: > e.g. observer is created with paths (relative path to root), and callback is invoked with path that has change

Note that there is no single "root" - there is a root for each Bucket File System (and local file picker selection) [^1]. 
So an observer created via a path would also need to know what that path is relative to, which is how we end up needing a `FileSystemHandle`

That being said, I do see the appeal of creating an observer to some path relative to some other directory handle. In the past there have been concerns about similar path-based APIs because we didn't want to specify path separators (see https://github.com/WICG/file-system-access/issues/284) but that's not as much an issue for the Bucket File System. If this is an important use case, we could consider adding it to the API

---
[^1]: This is more or less implied by the spec right now, and I tried to make this explicit in #131 but that PR has since lost steam. I still think it's valid and worth merging! It is also holding up https://github.com/whatwg/fs/pull/10#issuecomment-1637326389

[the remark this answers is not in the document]
a-sully: Yeah, exactly 🫤 

I'll resolve this thread but please feel free to re-open if there's more to discuss. Thanks!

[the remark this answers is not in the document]
szewai: hmm I don't see how the original question is resolved. The thread is about replacing `FileSystemObserver` with change event listener / `observeOptions` to `FileSystemHandle`. Did you agree to make that change?
(My previous comments is that `FileSystemObserver` is needed **if** we can decouple it with `FileSystemHandle`, if we cannot, I don't see the reason to have a separate interface.)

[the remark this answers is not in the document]
a-sully: I just remembered (sorry - it's been nearly two years since I wrote the explainer) that I'd considered an event-based pattern when designing the API, but we settled on the observer pattern for the reasons described here: https://github.com/whatwg/fs/blob/main/proposals/FileSystemObserver.md#alternatives-to-the-observer-pattern

Those reasons still hold, so I think we should stick with `FileSystemObserver`

I'll also note that, from a practical perspective, `FileSystemObserver` has already shipped in Chromium browsers. I filed https://github.com/WebKit/standards-positions/issues/291 nearly a year and a half ago but did not get a response. It is much harder to make drastic changes like the one you're proposing here once an API has shipped and is being actively used.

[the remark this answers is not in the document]
szewai: Thanks for the pointer; I've missed that part. I am not sure I understand the reasons of Observer pattern is better than event target:

> - The callback may be [triggered recursively](https://w3ctag.github.io/design-principles/#guard-against-recursion) if a file is modified in response to a file system change

Do you mean if the callback itself modifies a file? I think the operation will be async so another event will be dispatched asynchronously. So there should be no recursion

> - An Observer pattern allows the user agent to batch changes if necessary. For example, during a "git pull" operation that causes a flood of file system changes, these changes could be batched rather than firing the callback dozens of times

We don't need to dispatch one event for each change; we can still batch the changes in one event (e.g. making `FileSystemChangeEvent` contains an array of `FileSystemChangeRecord`), like how you implement batching now I guess.

> I'll also note that, from a practical perspective, FileSystemObserver has already shipped in Chromium browsers. I filed https://github.com/WebKit/standards-positions/issues/291 nearly a year and a half ago but did not get a response. It is much harder to make drastic changes like the one you're proposing here once an API has shipped and is being actively used.

Yeah apologies for the late response! I missed the thread until Anne pinged me about it recently. I appreciate that you are able to move things forward. I think we are positive about the API, but have concerns on the interfaces (the complicated Observe pattern). Though Chromium has shipped the API, it looks the API is still marked as [non-standard](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemObserver) and subject to change. Given the [current usage data](https://chromestatus.com/metrics/feature/timeline/popularity/5021), the compatibility risk should be pretty low, so I figure it might be still good timing to suggest big changes.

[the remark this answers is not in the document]
a-sully: > > * The callback may be [triggered recursively](https://w3ctag.github.io/design-principles/#guard-against-recursion) if a file is modified in response to a file system change
> 
> Do you mean if the callback itself modifies a file? I think the operation will be async so another event will be dispatched asynchronously. So there should be no recursion

Yeah, the callback may modify a file, though because of the sync methods of `FileSystemSyncAccessHandle`, the modifying operation (which triggers another callback/event) may be sync [^1]

That being said, batching events could help blunt the effects of this recursion so maybe that's not as big of an issue as I'd originally thought? Implementations would _have_ to batch changes generated from these sync operations to avoid excessive recursion

[^1]: For what it's worth the rationale for signaling events in response to `FileSystemSyncAccessHandle`'s sync modifying operations (`write()` and `truncate()`) is discussed in https://github.com/whatwg/fs/blob/main/proposals/FileSystemObserver.md#signaling-changes-made-via-a-filesystemsyncaccesshandle (as opposed to the discarded alternative https://github.com/whatwg/fs/blob/main/proposals/FileSystemObserver.md#signal-changes-made-via-a-filesystemsyncaccesshandle-only-on-close)

> > * An Observer pattern allows the user agent to batch changes if necessary. For example, during a "git pull" operation that causes a flood of file system changes, these changes could be batched rather than firing the callback dozens of times
> 
> We don't need to dispatch one event for each change; we can still batch the changes in one event (e.g. making `FileSystemChangeEvent` contains an array of `FileSystemChangeRecord`), like how you implement batching now I guess.

One complication with batching events on `FileSystemHandle` is that the object we're queueing to is mutable[^2]. e.g. if you queue changes to a `FileSystemDirectoryHandle` which is then moved, at which path (and therefore `FileSystemHandle`) do you report the changes happened? As I mentioned early on in this thread, I find the idea that an observer can be moved to an unintuitive implication of hanging `observeOptions` from a `FileSystemHandle` (along with that two `FileSystemHandle`s are equal but report different observing statuses), but ultimately this is something for which we could specify well-defined behavior

One reason for wanting `FileSystemObserver` was to establish a precedent that an observer watches an unchanging path[^3] which is independent from the `FileSystemHandle` once the observer is created. My intention was that would match the semantics of a follow-up proposal which would allow watching a path without requiring a live `FileSystemHandle`: see https://github.com/WICG/file-system-access/pull/457

So there could be a world in which we scrap `FileSystemObserver` in favor of `FileSystemHandle.observeOptions`, but we'd have to be okay with the inconsistency that:

1. `directoryHandle.startObserving()` (API TBD) followed by `directoryHandle.move()` _would move_ the observer to the new location (since it's tied to `directoryHandle` itself)
2. `registration.fileSystem.subscribe(directoryHandle)` followed by `directoryHandle.move()` _would NOT move_ the subscription to the new location (since it was configured by `directoryHandle`, but is thereafter independent from it)

That being said, that follow-up proposal also requires introduction of the same `FileSystemChangeEvent` type you're proposing in this thread. The reason that proposal uses the Event pattern rather than the Observer pattern is due to necessity - it must conform to the requirements of an API that can send events to wake up Service Workers. I still think the Observer pattern is a more natural fit, but we don't really have a choice. Changing _this_ proposal to use `observeOptions` could allow for re-use of that `FileSystemChangeEvent` type

> Yeah apologies for the late response! I missed the thread until Anne pinged me about it recently. I appreciate that you are able to move things forward. I think we are positive about the API, but have concerns on the interfaces (the complicated Observe pattern). Though Chromium has shipped the API, it looks the API is still marked as [non-standard](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemObserver) and subject to change. Given the [current usage data](https://chromestatus.com/metrics/feature/timeline/popularity/5021), the compatibility risk should be pretty low, so I figure it might be still good timing to suggest big changes.

I do appreciate the feedback, even if I wish it came earlier! 😛

Overall I'm not personally _opposed_ to the `observeOptions` proposal, but we should be clear about the tradeoffs that entails. To summarize the discussion so far (please let me know if I'm missing anything else):

Pros:

- Avoid introducing a new API surface (`FileSystemObserver`)
- Potential re-use of `FileSystemChangeEvent`

Cons:

- Implementations must batch synchronous changes to avoid recursive events (though I'd recommend doing this for the Observer pattern, too)
- Equivalent handles report different watch states - e.g. `a.isSameEntry(b)` may be true but `a` reports "observing" while `b` reports "not observing". Establishes a precedent that handles are not interchangeable
- Various unintuitive (I would argue) behaviors when a `FileSystemHandle` moves
- Chromium has already shipped `FileSystemObserver`, so the bar for making breaking changes is high (though it's still "non-standard", so not unclearably high)

@nathanmemmott @mingyc @fergald any thoughts here?

[^2]: There's an alternative timeline in which `FileSystemHandle.move()` does not modify the handle itself but instead _returns a new `FileSystemHandle` pointing to the new location_ but alas it's too late for that

[^3]: Another consideration was trying to match the behavior of native APIs for observing file system changes. The most modern is [FSEvents on MacOS](https://developer.apple.com/documentation/coreservices/file_system_events) - an [`FSEventStream` is configured with `pathsToWatch`](https://developer.apple.com/documentation/coreservices/1443980-fseventstreamcreate). On Linux, `inotify` also [creates watches with a `pathname`](https://man7.org/linux/man-pages/man2/inotify_add_watch.2.html), though that `pathname` is immediate converted into an inode, so this requires [shenanigans](https://crsrc.org/c/content/browser/file_system_access/file_path_watcher/file_path_watcher_inotify.cc;l=807-808;drc=56d30b6435d4bb9430d77af7a2b441372426d486) to update the observer if the path of the inode changes. If an observer was to instead move along with a `FileSystemHandle`, similar shenanigans would be needed when using FSEvents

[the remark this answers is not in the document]
a-sully: > * Chromium has already shipped `FileSystemObserver`, so the bar for making breaking changes is high (though it's still "non-standard", so not unclearably high)

I'd be curious if Blink API owners have opinions on this point specifically. @domenic any thoughts? See also the [I2S](https://groups.google.com/a/chromium.org/g/blink-dev/c/6oOaFmia2dc/m/aSEbVTABBQAJ) and https://github.com/w3ctag/design-reviews/issues/868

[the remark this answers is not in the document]
domenic: In Blink, one of our top commitments is [to compatibility](https://docs.google.com/document/d/1RC-pBBvsazYfCNNUSkPqAVpSpNJ96U8trhNkfV0v9fk/edit?tab=t.0). When developers ship code against an API, we owe them confidence that it will keep working. Revising a shipped API surface is thus a last resort.

It's for this reason we work hard to get wide review, e.g. via WebKit standards positions and W3C TAG review, or origin trial data, before shipping. This helps us get a signal that even if we are shipping first, and thus we haven't achieved formal "standard" status, we have as much indication as possible that the first-shipped API will achieve eventual interoperability. On this basis, the team made a commitment to the API owners in their Intent to Ship that they believed this would be a stable API for developers to use.

Of course, if a new version of the API is significantly better, then we can revisit. But from what I can tell, switching to the new proposal now would:

* Break existing sites and libraries that already depend on the shipped syntax.
* Undermine developers’ trust that early experimentation is safe.
* Create churn for developers without a clear, _user_-facing benefit.

Historically, the only time Blink has accepted these costs is when another engine ships an alternate design that demonstrably gains developer traction. So one possibility here is that if WebKit wants to prototype a different shape, we can watch adoption data for a release or two and reconvene. We would weigh this against the cited "strongly positive" developer sentiment for the current API that was written about in the Intent to Ship: if developers are even more positive about WebKit's API, then it could be worth taking on the churn. Until then, keeping the current API stable best serves the ecosystem.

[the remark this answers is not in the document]
szewai: Thanks for the detailed response! I agree most of your analysis and have some questions about Cons below.

> Cons:
> 
> Implementations must batch synchronous changes to avoid recursive events (though I'd recommend doing this for the Observer pattern, too)

To my understanding, the event handler interfaces will have the same behavior as the Observer ones like:
- Observe
    - Set `onchange` (callback) and  `observeOptions` on `FileSystemHandle`
    - Create `FileSystemObserver` with `FileSystemObserverCallback` and invoke `observe` with `FileSystemHandle`
- Unobserve
    - Set `observeOptions` on `FileSystemHandle` (e.g. to null to signal stopping observe)
    - Invoke `unobserve` on `FileSystemObserver` with `FileSystemHandle`
- When change happens
    - Dispatch `FileSystemChangeEvent` (that contains change information) to `FileSystemHandle`
    - Invoke `FileSystemObserverCallback` with `FileSystemChangeRecord ` (that contains observed `FileSystemHandle`)
So why does this not affect Observer pattern since the client can still write to file synchronously in `FileSystemObserverCallback`?

> Equivalent handles report different watch states - e.g. a.isSameEntry(b) may be true but a reports "observing" while b reports "not observing". Establishes a precedent that handles are not interchangeable

I think two handles should be "interchangeable” for performing file operations, but not sure why they need to be interchangeable for registering notifications. Normally in Observer pattern, the Observer would keep the observed target alive, so it seems technically the observed handle will have “state” that “equivalent” handle does not have.

> Various unintuitive (I would argue) behaviors when a FileSystemHandle moves

I wonder how the Observer pattern can avoid the confusion when an observed `FileSystemHandle` is moved. For example, if handle A is created with "old.txt", registered to observer, gets moved to "new.txt" and "old.txt" is created again; for the create operation, does observer invoke callback with A as root? Maybe we should introduce a `ConstFileSystemHandle` to better represent a path?

> Chromium has already shipped FileSystemObserver, so the bar for making breaking changes is high (though it's still "non-standard", so not unclearably high)

This might be the biggest concern since it could impact your current users. We don’t think the interface is a blocker, but it seems not optimal for usability and future development (Pros above), so we’d like to see if we can work together on a cleaner solution before this gets finalized (PR is merged and it becomes standard). 

Given the current usage data of this API (0.000060% of page loads), shipping a different design does not seem necessary to decide which design is better (I suspect the interoperability issue that “I need to use different APIs in different browsers” will likely get more attention than “I think this design is better”), we may just ask for input from developers or vendors in this thread (like @asutherland from Mozilla, or developers outside of Chromium you have discussed with), with the listed Pros and Cons.

PR #166, opened 2024-12-19
Update non-normative text for DirectoryHandle's removeEntry()

rahulsingh-msft:
This change updates only the non-normative section for
FileSystemDirectoryHandle's removeEntry() method.

It clarifies that the Promise is rejected with a "NotFoundError"
DOMException when an entry corresponding to the `name` parameter does
not exist.

Further, it adds that the Promise is rejected with an
"InvalidModificationError" DOMException when attempting to remove a
non-empty directory with the `recursive` option set to false.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -980,9 +980,9 @@ The <dfn method for=FileSystemDirectoryHandle>getDirectoryHandle(|name|, |option
      directory named |name|, this will attempt to delete that file or directory.
 
      Attempting to delete a file or directory that does not exist results in a promise
-     being rejected with a "NotFoundError" DOMException. Similarly, attempting to delete
-     a non-empty directory results in a promise being rejected with an
-     "InvalidModificationError" DOMException.
+     being rejected with a "NotFoundError" DOMException. Attempting to delete a non-empty
+     directory results in a promise being rejected with an "InvalidModificationError"
+     DOMException.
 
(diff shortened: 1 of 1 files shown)


annevk:
@rahulsingh-msft you need to work with someone at Microsoft (possibly @diekus) to get https://github.com/whatwg/participant-data updated as Microsoft currently hasn't signed up for the Storage Workstream (see also https://whatwg.org/workstreams).

index.bs
```diff
@@ -979,15 +979,18 @@ The <dfn method for=FileSystemDirectoryHandle>getDirectoryHandle(|name|, |option
...
+     DOMException.
 
   : await |directoryHandle| . {{FileSystemDirectoryHandle/removeEntry()|removeEntry}}(|name|, { {{FileSystemRemoveOptions/recursive}}: true })
   :: Removes the [=/file system entry=] named |name| in the [=directory entry=]
      [=locate an entry|locatable=] by |directoryHandle|'s [=FileSystemHandle/locator=].
      If that entry is a directory, its contents will also be deleted recursively.
 
-     Attempting to delete a file or directory that does not exist is considered success.
+     Attempting to delete a file or directory that does not exist results in a promise
+     being rejected with a "NotFoundError" DOMException.
```
annevk, line 993: This change seems to presuppose an outcome to #11. The normative text doesn't seem to currently reject at all.

rahulsingh-msft: Thank you for taking a first look @annevk! I'll pick this back up after the holidays. Will work with @diekus to get the Participant Data updated in the meantime.

